PR #3, opened 2019-03-19
Remove tiangolo.com from the CORS origin list

joshua-s:
This should not be enabled on every project unless tiangolo.com provides functionality necessary for the project to work.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -10,7 +10,7 @@
     "secret_key": "changethis",
     "first_superuser": "admin@{{cookiecutter.domain_main}}",
     "first_superuser_password": "changethis",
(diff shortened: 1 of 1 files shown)


tiangolo:
Yes, it provides functionatily for development.

http://localhost.tiangolo.com points to your `localhost` (`127.0.0.1`)
http://local.dockertoolbox.tiangolo.com points to the default Docker Toolbox IP if you have Docker Toolbox installed (`192.168.99.100`)

Both are local IPs.

Having a domain pointing to them allows debugging certain things that are not easy to debug otherwise (like CORS configurations).

joshua-s:
Thanks for clarifying! I should have looked deeper in the docs.

tiangolo:
No problem, thanks for coming back and closing it :+1:

tabuchid:
@tiangolo Just out of curiosity. How have you setup those domains to point at localhost? I recently was trying to do something similar using route53 and couldn't get it to work.

PR #4, opened 2019-03-24
Fixed read_user_by_id docstring

mpclarkson:
Changed doctring from "Get a specific user by username (email)" to "Get a specific user by id". Not the biggest PR in the world ;)


tiangolo:
Thanks! Good catch.

PR #5, opened 2019-03-29
Update User Schema

mmcloud:
Changed User attribute to EmailStr from str. Change Inheritance of UserInCreate and UserInUpdate to UserBase from UserBaseInDB to remove id input.


tiangolo:
This would require installing `email-validator`. But it might be cumbersome during development and testing to have to create real valid emails.

It might be worthwhile for some cases as an addition/upgrade, but might be cumbersome for some developers to have it by default.

PR #6, opened 2019-04-11
Fix frontend hijacking /docs in development

tiangolo:
:bug: Fix frontend hijacking /docs in development. Using latest https://github.com/tiangolo/node-frontend with custom Nginx configs in frontend.


mushmore123 (commented):
Gg

PR #9, opened 2019-04-12
removed postgres_password from alembic.ini

ebreton:
read it from env var instead

This allow to keep all environment variables in the *env files.


tiangolo:
This looks great, I'll review it soon.

tiangolo:
Thanks! Great job!

I just updated the URL string generation to use f-strings.

Sorry I ended up merging the latest changes into your branch :grin:

PR #17, opened 2019-04-27
Feature to forward CLI args to pytest

ebreton:
Minor (but useful) modification to pass CLI arguments to pytest (when needed)

Default:
```
docker-compose exec backend-tests /tests-start.sh
```

Stop on first error:
```
docker-compose exec backend-tests /tests-start.sh -x
```

Stop on first error, and filter and run debugger:
```
docker-compose exec backend-tests /tests-start.sh -x -k my-filter --pdb
```

> because test should be easy 🎈

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,8 @@ After using this generator, your new project (the directory created) will contai
 
 ### Next release
 
+* Update development scripts.
+
 ### 0.3.0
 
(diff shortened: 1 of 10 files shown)


tiangolo:
Awesome! Thanks :tada: :clap:

PR #18, opened 2019-04-27
Update pipfile for python37

ebreton:
Pipfile was left behind 😭 

I have also added autopep8 (which can prove useful)


tiangolo:
I have been using `black` for all the code, but as it doesn't support Python 3.7 yet, I have had to keep the local `Pipfile` with Python 3.6 :disappointed:

ebreton:
fair enough :)

PR #19, opened 2019-04-27
Fix the paths of the scripts in README

ebreton:
a fix on my way...


tiangolo:
Oops! Good catch. Thanks!

PR #20, opened 2019-04-27
Feature reusable normal user for tests

ebreton:
The main feature is the added fixture `normaluser_token_headers` which goes as a companion for the existing fixture `superuser_token_headers`.

The fixture does not create a brand new user everytime it is called, but the same user with its email defined from `config.EMAIL_TEST_USER`. This allows to limit the number of random users created in the database, and provides a control on (at least) one normal user.

Behind the hoods, the password for this 'pre-defined' user is reseted every time it is needed, which keeps the security up to the standards (no password needs to be defined anywhere for this particular test configuration)

As a side bonus, the function `byemail_authentication_token(email)` can be used to get a valid authorization token for the sole given argument of an email (changing the password on its way though)


tiangolo:
Thanks! I'll review it soon.

wassname:
Thanks for this. I'm not suggesting you do it this way, but another possibility to acheive this is to use pytest fixtures and yield to clean up the created users:

```py
@pytest.fixture(scope="module")
def random_user():
    password = faker.password()
    user_in = UserCreate(username=faker.name(), email=faker.email(), password=password)
    user = crud.user.create(
        db_session=db_session, user_in=user_in
    )
    # Attach the fake password for testing purposes
    user.password = password
    yield user
    # clean up fake user after tests pass or fail
    crud.user.remove(db_session=db_session, id=user.id)
```

tiangolo:
Thank you @ebreton for your contribution! :rocket: :tada: :cake:

PR #23, opened 2019-05-03
Simplify CRUD definitions, make a clearer distinction between schemas and models

ebreton:
This PR facilitates the creation of the basic CRUD operations for new db_models objects.

**Motivation: get DRYer**

If you don't care for IDE embedded auto-completion, you willl not need to copy/paste `app.crud.items.py` into `app.crud.your_model.py` file anymore. One line is sufficient in `app.crud.__init__.py` (with appropriate import):

```
your_model = CrudBase(YourModel)
```

If you care about "intellisense", then you will still have a bit of copy-pasting for the sake of auto-completion, and leverage your favorite IDE. But you will still gain time (and lines of code)

**Solution**
The PR therefore defines a `CrudBase` class, which instances will provide the get/get_multi/create/udpate/delete/remove operations for a given SQLAlchemy object. I have tried to be as detailed as possible in the documentation of `app.crud.base.py`

Along with this new tool, I have added a new `SubItem` model that showcases the usage of `CrudBase`. I have also modified Item with the same objective. Tests have been added or updated consequently.

> because we should be DRY 🚀 

**Bonus 1: reflect the fastapi documentation by using `schemas` (when pydantic) and `models` (when sqlalchemy)**

I have updated the tree structure to reflect the latest documentation from fastapi (https://fastapi.tiangolo.com/tutorial/sql-databases/). Hence renaming `models` to `schemas` and `db_models` to `models`.

**Bonus 2: run the test more easily**

when runing `sh tests.sh` from the root directory, arguments are forward to `scripts/test.sh` and then to `.../test-start.sh`. 

Thats allows the user to pass whatever argument he would pass to pytest, like `-x -v -k filter --pdb` (pick your favorite)

when `sh tests.sh` fails and you want to quicky run the test again and again (**without** deleting the entire `testing-project`and generating again (including docker image), then you can user `sh test-again.sh` that will rsync the testing-project src code with your latest modifications.

**Bonus 3: easy casting from schema to model (and vice versa)**

Not only we I have tried to be very explicit on which kind object is used, i.e

```
from app.models import subitem as models_subitem
from app.schemas import subitem as schemas_subitem

subitem = CrudBase(models_subitem.SubItem, schemas_subitem.SubItem)
```

But the models provide 2 functions to ease the casting_

```
class CustomBase(object):

    def to_schema(self, schema_cls):
        return schema_cls(**self.__dict__)

    @classmethod
    def from_schema(cls, schema_obj):
        return cls(**jsonable_encoder(schema_obj))
```

which allows you to easily cast a model instance to any schema you provide, or create the model object from a schema instance.

Those 2 functions are used in the base crud definitions of `update` and `create`, and makes sure that all types will be properly cast (in particular non directly JSONable ones like datetime or enums)

**Bonus 4: normal user for tests**

without the need to create a new one in each and every tests...

That basically comes from PR #20


tiangolo:
Very interesting ideas! I'll check it thoroughly soon.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,161 @@
...
+        queried = db_session.query(self.db_model).filter(self.db_model.id == obj_id)
+        counted = queried.count()
+        if counted > 0:
+            queried.delete()
+            db_session.commit()
+        return counted
+
+    def remove(self, db_session: Session, *, obj_id: int) -> Optional[Base]:
+        """
+        remove does the same job as delete, with a different return valie
```
wassname: Typo here

wassname:
Just to chime in, I'm not a contributor but been using the contents of this PR in my project and it works fine and seems cleaner.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,161 @@
...
+        Keyword Arguments:
+            skip {int} -- Number of rows to skip from the results (default: {0})
+            limit {int} -- Maximum number of rows to return (default: {100})
+        
+        Returns:
+            List[Optional[Base]] -- Array of DB instances according given parameters. Might be empty if no objets are found.
+        """  # noqa
+        return db_session.query(self.db_model).offset(skip).limit(limit).all()
+
+    def create(self, db_session: Session, *, obj_in: BaseModel) -> Base:
```
wassname: After using this, I noticed you could make it more general by passing in extra arguments for db_model.

```py
def create(self, db_session: Session, *, obj_in: BaseModel, **kwargs) -> Base:
    obj_in_data = jsonable_encoder(obj_in)
    obj = self.db_model(**obj_in_data, **kwargs)
    db_session.add(obj)
    db_session.commit()
    db_session.refresh(obj)
    return obj
```

Commenly parent_id, child_id, and similar database only args would go in there.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+            crud_item = CrudItem(Item)
+
+        Arguments:
+            db_model {Base} -- Class of the DB model which CRUD methods will be provided for
+        """  # noqa
+        self.db_model = db_model
+
+    def get(self, db_session: Session, obj_id: int) -> Optional[Base]:
```
wassname: Here's how to add a get_by method that will let you pass in arbitraty attributes:

```py
    def get_by(self, db_session: Session, **kwargs: Any) -> Optional[Base]:
        filter_kwargs = {getattr(self.db_model, k)==v for k, v in kwargs.items()}
        return db_session.query(self.db_model).filter(*filter_kwargs).first()
```

{{cookiecutter.project_slug}}/backend/app/app/crud/item.py
```diff
@@ -5,51 +5,31 @@
...
 
+    def create(self, db_session: Session, *, item_in: ItemCreate, owner_id: int) -> Item:
+        item_in_data = jsonable_encoder(item_in)
+        item = Item(**item_in_data, owner_id=owner_id)
+        db_session.add(item)
+        db_session.commit()
+        db_session.refresh(item)
+        return item
 
-def get_multi_by_owner(
```
wassname: I think we are now missing get_multi_by_owner, but it's still used in the api. So even though it doesn't show up in tests, it might be worth adding.

wassname:
Sorry for all the comment @ebreton, they are because I liked an used the PR :)

ebreton:
no worries @wassname , its awesome to see you comment and improve :) 
I am just deep underwater right now, but I appreciate 👍

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: indeed :)

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: I think any other attributes you might want to pass at creation time must be defined in your schema, and will be passed through `obj_in.

That is also the case with relationships like parent_id and child_id.

Let's say your User belongs to a city. The schemas would look like this:

```python
# Shared properties
class UserBase(BaseModel):
    email: str = None
    is_active: bool = True
    is_superuser: bool = False
    full_name: str = None
    city_id: int = None

    class Config:
        orm_mode = True


# Additional properties stored in DB
class UserBaseInDB(UserBase):
    id: int = None
    created_at: Optional[datetime] = None
    updated_at: Optional[datetime] = None


# Properties to receive via API on creation
class UserCreate(UserBaseInDB):
    email: str
    password: str
    created_at: datetime = Schema(datetime.utcnow())


# Properties to receive via API on update
class UserUpdate(UserBaseInDB):
    password: str = None
    updated_at: datetime = Schema(datetime.utcnow())
```

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: The idea looks good but I am afraid I am not up to the latest trend: I don't understand what `filter_kwargs`is ...

it looks like a set of booleans... but that does not make sense does it ?

Could you explain or give me a few pointers to the doc please 😃

{{cookiecutter.project_slug}}/backend/app/app/crud/item.py
ebreton: you are entirely right. nice catch 👍

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
wassname: SQLAchemy uses statements like `Model.field==value` to generate an object, that can be passed as an argument to perform filtering. So, it's a bit misleading but when you see:

`db_session.query(db_model).filter(db_model.field==value).first()`

It's the same as 
```
filter_args=(db_model.field==value)
db_session.query(db_model).filter(filter_args).first()
```

```
filter_args=(getattr(db_model, 'field')==value)
db_session.query(db_model).filter(filter_args).first()
```

On looking at my code again, it would be better as:
```
def get_by(self, db_session: Session, **kwargs: Any) -> Optional[Base]:
        filter_args = [getattr(self.db_model, k)==v for k, v in kwargs.items()]
        return db_session.query(self.db_model).filter(*filter_args).first()
```

And if it's too confusing maybe it's better not to use it in this example project anyway!

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
wassname: Oh I see, that makes sense thanks.

wassname:
A little bit tangential, but on the topic of crud, [here's an approach I came across using sqlalchemy mixins](https://github.com/absent1706/sqlalchemy-mixins/blob/master/sqlalchemy_mixins/activerecord.py).

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: Thanks a lot for the explanation :pray: Not used to SQL alchemy syntax yet.

I think the comprehension list with the array is simpler to understand. I will go with it.

I am in the process or updating this PR with the ORM feature added lately. I will integrate your suggestions in the process 😄

ebreton:
> A little bit tangential, but on the topic of crud, [here's an approach I came across using sqlalchemy mixins](https://github.com/absent1706/sqlalchemy-mixins/blob/master/sqlalchemy_mixins/activerecord.py).

I will need to dive a bit further into SQLAlchemy, I have not used their MixIns yet

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
wassname: Thanks!, I appreciate the work involved in that

omrihar:
I've been using `sqlalchemy_mixins` for a while now, and actually it made me make some substantial changes to the template. For example, it's not necessary to pass the database as a dependency injection anymore, because you can set the session on the base class. It's quite convenient I must say. I still use the crud object though because sometimes the operations necessary can be quite complex.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
wassname: I came across a simpler way, you don't need to change it or anything, just sharing a simpler solution.
```
def get_by(self, db_session: Session, **kwargs: Any) -> Optional[Base]:
        return db_session.query(self.db_model).filter_by(kwargs).first()
```

wassname:
I don't suppose you could share an example, or a gist few anonymized files? It's always interesting to see implementation details of something we're working on.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: I have added `get_first_by` and `get_multi_by` based on your suggestion 🍰

{{cookiecutter.project_slug}}/backend/app/app/crud/item.py
ebreton: Added back, making use of `get_multi_by` from `CrudBase`

ebreton:
@tiangolo , I have put a lot of things in this one, which goes along with the initial purpose: simplify the crud definitions, and make a clearer distinction between schemas and models.

Hope you will find it useful... 🤞

wassname:
I like the models=>schemas, cleaning up the user crud. I've been using UserExpanded schemas to returns lists of items too. <3 pytest fixtures. 

Thanks, I've made wanted to make similar changes to my project.

ebreton:
Hi @tiangolo !

I was wondering if I could something to help you accept this PR 😇

thiras:
This PR looks amazing. Would be great if we could get merge.

{{cookiecutter.project_slug}}/backend/app/app/schemas/user.py
```diff
@@ -1,40 +1,49 @@
...
+from pydantic import BaseModel, Schema
 
 
 # Shared properties
 class UserBase(BaseModel):
-    email: Optional[str] = None
-    full_name: Optional[str] = None
-    is_active: Optional[bool] = True
-    is_superuser: Optional[bool] = False
+    email:str = None
```
HymanZHAN: Not so sure about removing all the `Optional`... Technically, a `str` type cannot have `None` as the value, right? Plus this does not work well with mypy.

ebreton: Hi @HymanZHAN ,

I had some issues with the Optionnal[] on native types (int, bool, str, ...), but I am not able to remember which ones exactly... but that is the reason why I removed them.

`None` should not be considered as the value of the string here, but more an indication that it is not set (i.e optionnal) but I agree with you that it is not as pretty as the original Optionnal[]

I will have to digg into it for the source pb :sweat_drops:

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,218 @@
...
+                    return super(CrudItem, self).create(db_session, obj_in=obj_in)
+
+                def update(self, db_session: Session, *, obj: models_item.Item, obj_in: schemas_item.ItemUpdate) -> models_item.Item:
+                    return super(CrudItem, self).update(db_session, obj=obj, obj_in=obj_in)
+
+
+            crud_item = CrudItem(models_item.Item, schemas_item.Item)
+
+        Arguments:
+            model {Base} -- Class of the DB model which CRUD methods will be provided for
```
fj-sanchez: Missing `schema` argument.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,218 @@
...
+from fastapi.encoders import jsonable_encoder
+from sqlalchemy.orm import Session
+
+from app.db.base_class import Base
+from pydantic import BaseModel
+
+
+class CrudBase:
+
+    def __init__(self, model: Base, schema: BaseModel):
```
fj-sanchez: I think it would be correct here to use `Type[BaseModel]` but I'm no 100% sure.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
```diff
@@ -0,0 +1,218 @@
...
+        
+        Arguments:
+            db_session {Session} -- Dependency injection of the Database session, which will be used to commit/rollback changes.
+        
+        Keyword Arguments:
+            skip {int} -- Number of rows to skip from the results (default: {0})
+            limit {int} -- Maximum number of rows to return (default: {100})
+        
+        Returns:
+            List[Optional[Base]] -- Array of DB instances according given parameters. Might be empty if no objets are found.
```
fj-sanchez: Typo in "objets". There are a few others, I'd suggest a spell-checker.

{{cookiecutter.project_slug}}/backend/app/app/crud/__init__.py
```diff
@@ -1 +1,11 @@
...
+from .item import item
+
+# For a new basic set of CRUD operations, on a new object, let's say 'SubItem',
+# you could also simply add the following lines:
+
+from .base import CrudBase
+from app.models import subitem as models_subitem
+from app.schemas import subitem as schemas_subitem
+
+subitem = CrudBase(models_subitem.SubItem, schemas_subitem.SubItem)
```
fj-sanchez: Is it a good idea to use module variables (effectively globals/singletons)? What happens if you get multiple requests and one of them is using this object?

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: Thanks for the suggestion. spell-checker installed 💯

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: 👍

{{cookiecutter.project_slug}}/backend/app/app/crud/__init__.py
ebreton: Those variables are just "gateways" from/into the DB. It will be the DB which will make sure that there is no conflict when multiple requests come in.

{{cookiecutter.project_slug}}/backend/app/app/crud/base.py
ebreton: not sure either... I will have to take a closer look at this one (or get feedback from others ;) )

tiangolo:
Thank you @ebreton for your contribution! :cake: 

Thanks everyone for the PR feedback.

@omrihar I suggest you don't add the session to your models instead of passing the DB session around. That would mean that you would have to resort to SQLAlchemy's thread-local handling (or the equivalent from some other library). Those techniques assume a framework that assigns exactly one single thread per request. But as FastAPI is async, a single thread could handle more than one request concurrently, and those models using the same session would create conflicts and errors. Also, as FastAPI could run some parts of the code in a threadpool (normal `def` functions), those sessions that were local to one thread would not be available for those other functions.

If you really want to use a technique like those, you would have to use `contextvars`, but you would have to use Python 3.7, make sure you set the ContextVar in a middleware, and open and close the sessions in a dependency with yield, etc. It's possible, but it might be a bit fragile. And as it works kinda "magically", understanding it while implementing it is quite mind-bending. For more details about it, see the section about Peewee, that does that same thing, more or less: https://fastapi.tiangolo.com/advanced/sql-databases-peewee/

Anyway, I suggest you just keep the session around. It's a lot more explicit and easy to understand and debug.

---

I liked several of the ideas here @ebreton ! :rocket: 

It was a bit difficult to review as there were a lot of changes that were not related to the PR, so it took me a while to get around it :sweat_smile:. But I just finished updating it to leave it in a state where it's just the proposed change.

I updated the `CRUDBase` to simplify it and use Generics, that way the other sub-classes can still have all the types and editor support without having to override all the methods.

I want to have just the minimum code that can work as a base for almost any project. So I removed several things that aren't necessarily always needed. I removed the sub-items models, and some methods. I removed the custom logs, as logging personal information that ends up stored in logs somewhere would make it difficult to be conformant with GDPR and similar privacy regulations. I'm not sure about adopting a specific format for docstrings, but I prefer to delay that decision until later and merge the main idea here.

Any other of the changes can be implemented/reviewed in independent PRs, it's a lot easier like that. And that way we can have the `CRUDBase` here now.

Thanks for your effort! :taco: :tada:

ebreton:
Thanks @tiangolo , that's awesome !!

I apologize for the heaviness of the PR, and I will surely make the other PR smaller in the future. 🔬 😄 

Thank you so much for fastAPI and this sibling project. We are using it on daily basis 🌟 💎

timhughes:
@ebreton @tiangolo Just a thought on this Pull Request as someone who is learning. This PR has taken the example application away from the examples in the FastAPI  documentation. It would be good if the example application and the docs were somewhat similar so that people could see real world examples. Specifically I am talking about the CRUD changes and things like UserInDB which appears to be no longer in use. I do appreciate that the code is cleaner in it's current incantation (and I have learnt a thing ot two about generics - thanks) but for someone who is following the documentation and trying to work out how everything fits together it makes understanding more difficult. 

I am not overly concerned from my own point of view because  just checked out the revision before the merge and that helped me understand a lot better. I just think the the old style of code was simpler to understand for someone coming at this for the first time.

Thanks for all the awesome work.

ebreton:
You are right: I have focused on documenting the PR, and I have failed at updating the doc. 

I guess it will come, but any help from you would be more than welcomed 😇 . 

Would you please make a try at a PR where you could share your experience ? I would say that it would make the learning process much simpler, even simplest than the previous one once you get the trick (and one can always stick to the first version if it remains an option in the doc too)

Issue #28, opened 2019-05-20
[BUG] SQLAlchemy InvalidRequestError on relationships if one of the model is not yet imported


ebreton:
**Describe the bug**

The application crashes at start-up, when initializing data if :
1. a relationship is defined ...
2. ... with a model not already imported at the time of execution.

```bash
backend_1        | INFO:__main__:Starting call to '__main__.init', this is the 2nd time calling it.
backend_1        | INFO:__main__:Service finished initializing
backend_1        | INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
backend_1        | INFO  [alembic.runtime.migration] Will assume transactional DDL.
backend_1        | INFO  [alembic.runtime.migration] Running upgrade  -> d4867f3a4c0a, First revision
backend_1        | INFO  [alembic.runtime.migration] Running upgrade d4867f3a4c0a -> ea9cad5d9292, Added SubItem models
backend_1        | INFO:__main__:Creating initial data
backend_1        | Traceback (most recent call last):
backend_1        |   File "/usr/local/lib/python3.7/site-packages/sqlalchemy/ext/declarative/clsregistry.py", line 294, in __call__
backend_1        |     x = eval(self.arg, globals(), self._dict)
backend_1        |   File "<string>", line 1, in <module>
backend_1        | NameError: name 'SubItem' is not defined
...
backend_1        | sqlalchemy.exc.InvalidRequestError: When initializing mapper mapped class Item->item, expression 'SubItem' failed to locate a name ("name 'SubItem' is not defined"). If this is a class name, consider adding this relationship() to the <class 'app.db_models.item.Item'> class after both dependent classes have been defined.
base-project_backend_1 exited with code 1
```
**To Reproduce**

create a new db_models/subitems.py (could be copied from item.py)

It is important that this new model has a relationship to another one, e.g Item

```python
from sqlalchemy import Column, ForeignKey, Integer, String
from sqlalchemy.orm import relationship

from app.db.base_class import Base


class SubItem(Base):
    id = Column(Integer, primary_key=True, index=True)
    title = Column(String, index=True)
    description = Column(String, index=True)
    item_id = Column(Integer, ForeignKey("item.id"))
    item = relationship("Item", back_populates="subitems")
```

Adapt db_models/item.py with the new `relationship`

```python
...

class Item(Base):
    ...
    subitems = relationship("SubItem", back_populates="item")
```

Declare the new SubItem in db/base.py as per the documentation

[...]

PR #29, opened 2019-05-20
Fix sqlalchemy class lookup

ebreton:
This is a one-line fix to avoid the crash of the application in certain edge cases, where you defined an SQLAlchemy relationship where one of the model is not yet imported at the time the application creates it initial data.

See Issue #28 for all the details

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/app/app/db/init_db.py
+++ b/{{cookiecutter.project_slug}}/backend/app/app/db/init_db.py
@@ -2,6 +2,8 @@ from app import crud
 from app.core import config
 from app.models.user import UserCreate
 
+from .base import *  # noqa
+
 
 def init_db(db_session):
     # Tables should be created with Alembic migrations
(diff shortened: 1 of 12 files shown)


{{cookiecutter.project_slug}}/backend/app/app/db/init_db.py
```diff
@@ -2,6 +2,8 @@
 from app.core import config
 from app.models.user import UserCreate
 
+from .base import *  # noqa
```
ebreton: the `# noqa` here is important, others wise the autoformat tools (lint.sh) would remove this line

ebreton:
I will do that tomorrow :ok_hand:

ebreton:
Good morning @tiangolo , the `from app.db import base` works fine :notes: . Changed made, and comment added.

Let me know whether there is anything else. I was hesitating for instance to add a new SubItem db_models to get automated testing and avoid regression (as described in #28), but I didn't... it seemed like too much noise...

tiangolo:
Awesome! Thanks a lot for your work :tada: :cake:

PR #30, opened 2019-05-20
Use faker for tests

ebreton:
Because tests should be first citizens pieces of code... they deserve proper random (faked) data :school_satchel:

Faker is a new dependency, but only for tests (not included in backend dockerfile)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,10 @@ After using this generator, your new project (the directory created) will contai
 
 ### Next release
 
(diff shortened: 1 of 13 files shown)


tiangolo:
Thanks!

In fact, I initially used Faker (for the Flask version), but as currently the DB gets polluted with tests that are not removed afterwards, I started getting collisions frequently. And I realized that I didn't need real data, just random strings that I could check.

Let's leave this on standby until I update the tests to use Starlette's TestClient, and we can solve automatic rollbacks in tests.

ebreton:
Yes, sounds good. I will create another PR in due time 👍

tiangolo:
Thanks!

Issue #31, opened 2019-05-21
[BUG] DB OperationalError


ebreton:
**Describe the bug**

The application returns (randomly) 500 errors.

When looking at the logs, we can see some OperationError occuring: `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
`

Following the link at the end of the traceback, we have the following information from SQLAlchemy web site:

> Exception raised for errors that are related to the database’s operation and not necessarily under the control of the programmer, e.g. an unexpected disconnect occurs, the data source name is not found, a transaction could not be processed, a memory allocation error occurred during processing, etc.

> This error is a DBAPI Error and originates from the database driver (DBAPI), not SQLAlchemy itself.

> The OperationalError is the most common (but not the only) error class used by drivers in the context of the database connection being dropped, or not being able to connect to the database. For tips on how to deal with this, see the section Dealing with Disconnects.

Here is the full traceback

```bash
[2019-05-20 07:46:24 +0000] [12] [INFO] ('10.0.16.163', 56442) - "GET /api/v1/audit_app/audit_groups/me HTTP/1.1" 500
[2019-05-20 07:46:24 +0000] [12] [ERROR] Exception in ASGI application
Traceback (most recent call last):
File "/usr/local/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 372, in run_asgi
result = await asgi(self.receive, self.send)
File "/usr/local/lib/python3.7/site-packages/starlette/middleware/errors.py", line 125, in asgi
raise exc from None
File "/usr/local/lib/python3.7/site-packages/starlette/middleware/errors.py", line 103, in asgi
await asgi(receive, _send)
File "/usr/local/lib/python3.7/site-packages/starlette/middleware/base.py", line 27, in asgi
response = await self.dispatch_func(request, self.call_next)
File "/app/app/main.py", line 37, in db_session_middleware
response = await call_next(request)
File "/usr/local/lib/python3.7/site-packages/starlette/middleware/base.py", line 44, in call_next
task.result()
File "/usr/local/lib/python3.7/site-packages/starlette/middleware/base.py", line 37, in coro
await inner(request.receive, queue.put)
File "/usr/local/lib/python3.7/site-packages/starlette/exceptions.py", line 74, in app
raise exc from None
File "/usr/local/lib/python3.7/site-packages/starlette/exceptions.py", line 63, in app
await instance(receive, sender)
[...]

PR #32, opened 2019-05-21
Fix operation error

ebreton:
Details of the bug can be found in issue #31 

The solution makes use of the 'pessimistic' approach as described in https://docs.sqlalchemy.org/en/13/core/pooling.html#pool-disconnects

> The approach adds a small bit of overhead to the connection checkout process, however is otherwise the most simple and reliable approach to completely eliminating database errors due to stale pooled connections. The calling application does not need to be concerned about organizing operations to be able to recover from stale connections checked out from the pool

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,8 @@ After using this generator, your new project (the directory created) will contai
 
 ### Next release
 
+* Update development scripts.
+
 ### 0.3.0
 
 * PR <a href="https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/14" target="_blank">#14</a>:
--- a/README.md
+++ b/README.md
@@ -148,6 +148,10 @@ After using this generator, your new project (the directory created) will contai
 
 ### Next release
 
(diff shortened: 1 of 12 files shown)


tiangolo:
Thanks! That looks like a small change with a long time debugging and investigating :nerd_face: 

Were you able to test it? Did it solve the problem?

ebreton:
Hi @tiangolo , I do not have a reproductible test, yet here is what I have done:

- before fix, restart the `db` container, and query the API backend -> 500 errors come at once, in numbers

- after fix, restart the `db` container with same conditions -> there is a few-seconds lag in the first query (probably timeout and/or retrials from the connection pool), then a couple of 500 errors before the DB is up again

I have also checked the logs, and there are no more OperationalErrors after applying the fix (I will check again during the week)

From my point of view, the issue is fixed, but not automatically tested. Which does not really concern me, since the management of the connection cool is on the SQLAlchemy level, and therefore tested by them

tiangolo:
Cool!

Thanks for the explanation and the local tests you did. That's good enough :heavy_check_mark: 

Thanks for your work! :tada: :rocket: :cake:

PR #33, opened 2019-05-28
Update login.py

dmontagu:
Change `new_password` from a query parameter to a body parameter for security.

(Why this is problematic is discussed in the top answer to https://stackoverflow.com/questions/2629222/are-querystring-parameters-secure-in-https-http-ssl)


tiangolo:
Absolutely! :sweat_smile:

Thanks for fixing it! :tada: :rocket:

PR #40, opened 2019-06-13
Add email validation

kedod:
Add email validation in User model with pydantic EmailStr.

Btw have you considered using pipenv or requirements.txt files for python packages dependencies?
Installing packages one by one in dockerfiles is kind of ugly.

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/app/app/schemas/user.py
+++ b/{{cookiecutter.project_slug}}/backend/app/app/schemas/user.py
@@ -1,11 +1,11 @@
 from typing import Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, EmailStr
 
 
 # Shared properties
 class UserBase(BaseModel):
-    email: Optional[str] = None
+    email: Optional[EmailStr] = None
     is_active: Optional[bool] = True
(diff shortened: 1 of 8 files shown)


StephenBrown2:
@tiangolo Fixed in #23?

tiangolo:
Thanks for the discussion here everyone!

About Pipenv and installing in Dockerfiles, I'm still not completely happy with how these package dependency management tools work. For example, Pipenv is currently not very well maintained, I plan to switch to Poetry instead.

But also, installing in Dockerfiles lets you iterate quickly and take advantage of the Docker cache. For example, if you want to test a new Python package, you could add a line to your Dockerfile and build again, without having to wait for it to install everything from scratch.

But again, as I'm not entirely happy with all this yet, I don't want to "recommend" one way or another. So, for the moment, I suggest you update your local system to use the tools and workflow of your preference (`requirements.txt`, Pipenv, Poetry, etc.).

---

About this PR, yeah, #23 included some parts but I simplified it to have the minimum, as it covered a lot of different things in a single PR.

I just updated this PR to rebase it from the latest changes.

Thanks for your contribution @kedod ! :rocket: :tada: :cake:

PR #74, opened 2019-10-18
Fix typo in `read_item` GET view

jcaguirre89:
Swapped `read_user_me` for `read_item` for the single item get request. Looks like it was copied from the user paths and not updated. Doesn't change functionality but is more readable as it didn't make sense to have it called `read_user_me` and might be confusing. While I was at it I also updated the status code for not found to 404, so it's a more helpful error message if not found.


tiangolo:
Good catch! Thanks @jcaguirre89 ! :mag: :tada: :cake:

PR #77, opened 2019-11-06
:wrench: Remove tiangolo.com from default options

ayharano:
Last two entries from `backend_cors_origins` are related to tiangolo.com subdomains.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -10,7 +10,7 @@
     "secret_key": "changethis",
     "first_superuser": "admin@{{cookiecutter.domain_main}}",
     "first_superuser_password": "changethis",
-    "backend_cors_origins": "http://localhost, http://localhost:4200, http://localhost:3000, http://localhost:8080, https://localhost, https://localhost:4200, https://localhost:3000, https://localhost:8080, http://dev.{{cookiecutter.domain_main}}, https://{{cookiecutter.domain_staging}}, https://{{cookiecutter.domain_main}}, http://local.dockertoolbox.tiangolo.com, http://localhost.tiangolo.com",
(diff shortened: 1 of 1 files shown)


ayharano:
As this cookiecutter user, I am not sure why two of @tiangolo' subdomains would be considered into default choice, so this PR is intended to remove them from default.

dmontagu:
If you read the readme, he explains why they are in there — they point to domains that are useful during local development. If I recall correctly, it was something like some browsers don’t let you set cookies for IP addresses rather than actual domains. localhost.tiangolo.com points to 127.0.0.1 so that you can have an actual domain but still hit localhost. The docker one is similar, but for a different use case.

I don’t think these urls are *necessary* to include, but they have uses and weren’t just accidentally included. And since it’s so easy to remove them if you want to in your own projects, I personally don’t see the harm in keeping them.

ayharano:
I agree with the arguments that you wrote.

However, considering a _default_ option to populate in a project, it still sounds a bit off including template-filled values along hardcoded subdomains.

Considering among the options of
i. user modifying to remove them later after reading the README, or
ii. not including them by default, in a general case,

I would think that the later one sounds better.

tiangolo:
Thanks @ayharano for your suggestion. And, yep, the reason is what @dmontagu says.

The idea for this project generator is to be as simple to get started as possible, just as a baseline, to help developers start it and have "something" working right away.

To allow that, it's heavily opinionated in many aspects (e.g. why PostgreSQL and not MySQL or MariaDB?), but that way it provides something that "just works" from the start.

I expect developers to change their generated projects a lot, remove parts (e.g. Celery if they don't use it). The same way I expect them to remove those helper domains if they don't need them.

I'll close this PR now, but feel free to add more comments.

Issue #84, opened 2019-12-04
Backend not starting:  ImportError: cannot import name 'EmailStr' from 'pydantic.types'


merowinger92:
The backend is not starting due to an import error:

```
backend_1        | INFO: Stopping reloader process [1]
backend_1        | Checking for script in /app/prestart.sh
backend_1        | Running script /app/prestart.sh
backend_1        | INFO:__main__:Initializing service
backend_1        | INFO:__main__:Starting call to '__main__.init', this is the 1st time calling it.
backend_1        | INFO:__main__:Service finished initializing
backend_1        | INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
backend_1        | INFO  [alembic.runtime.migration] Will assume transactional DDL.
backend_1        | INFO:__main__:Creating initial data
backend_1        | INFO:__main__:Initial data created
backend_1        | INFO: Uvicorn running on http://0.0.0.0:80 (Press CTRL+C to quit)
backend_1        | INFO: Started reloader process [1]
backend_1        | Process SpawnProcess-1:
backend_1        | Traceback (most recent call last):
backend_1        |   File "/usr/local/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
backend_1        |     self.run()
backend_1        |   File "/usr/local/lib/python3.7/multiprocessing/process.py", line 99, in run
backend_1        |     self._target(*self._args, **self._kwargs)
backend_1        |   File "/usr/local/lib/python3.7/site-packages/uvicorn/supervisors/statreload.py", line 29, in handle_fds
backend_1        |     target(**kwargs)
backend_1        |   File "/usr/local/lib/python3.7/site-packages/uvicorn/main.py", line 307, in run
backend_1        |     loop.run_until_complete(self.serve(sockets=sockets))
backend_1        |   File "uvloop/loop.pyx", line 1417, in uvloop.loop.Loop.run_until_complete
backend_1        |   File "/usr/local/lib/python3.7/site-packages/uvicorn/main.py", line 314, in serve
backend_1        |     config.load()
backend_1        |   File "/usr/local/lib/python3.7/site-packages/uvicorn/config.py", line 186, in load
backend_1        |     self.loaded_app = import_from_string(self.app)
backend_1        |   File "/usr/local/lib/python3.7/site-packages/uvicorn/importer.py", line 23, in import_from_string
backend_1        |     raise exc from None
[...]

PR #85, opened 2019-12-04
Update utils.py so that pydantic>=1.2 is supported

merowinger92:
EmailStr changed from pydantic.types to pydantic.networks


merowinger92:
relates to issue #84

PR #86, opened 2019-12-04
Update user.py to support pydantic>=1.2  #84

merowinger92:
EmailStr was moved from pydantic.types to pydantic.networks


merowinger92:
relates to issue #84

tiangolo:
Thanks for coming back to close the issue! :+1:

Issue #97, opened 2020-01-07
Use . instead of source in shell scripts


gucharbon:
I'm using 18.04.3 LTS (Bionic Beaver) and python 3.7.5
I just created my own repository using cookiecutter.

I ran the script `./scripts/build-push.sh` and received the error below:

```
(venv) monsoon@-bash4.4:/c/Users/monsoon/fastapi/webdemo(21:19:57)$ ./scripts/build-push.sh
./scripts/build-push.sh: 6: ./scripts/build-push.sh: source: not found
```

After taking a look at [the template](https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/scripts/build-push.sh) I saw that `/bin/sh` is specified in shebang:

```sh
#! /usr/bin/env sh

# Exit in case of error
set -e
```

 Ubuntu uses `/bin/dash` in place of`/bin/sh` and not `/bin/bash`. [Ubuntu Wiki](https://wiki.ubuntu.com/DashAsBinSh#source) states that:
>The source builtin is a bashism (related checkbashisms warning text: 'source' instead of '.'). Write this simply as . instead. 

Isn't there anybody on Ubuntu or Debian which tried to run the scripts and had this error? I searched existing issue but found none.

Edit: I just saw in the README that user should use `bash`  to start scripts (as in the example: `TAG=prod FRONTEND_ENV=production bash ./scripts/build-push.sh`) but I still feel that this `.` should be used instead of `source` to be coherent with shebang and to allow usage of `TAG=prod FRONTEND_ENV=production ./scripts/build-push.sh`.

Anyway thanks for the awesome work :)

Shackelford-Arden:
@gucharbon While that documentation you provided does indicate the use of `dash` on Ubuntu based systems, if you run the command as shown in the shebang:

```shell
/usr/bin/env sh
```

On an Ubuntu system (tested on Ubuntu 18.04.4) you'll find that this works just fine.

As the documentation page you referenced indicates in the second paragraph from the top:

> The default login shell remains bash. Opening a terminal from the menu or shortcut [crtl-alt-t] provides interactive bash. A script run from the desktop or file manager, through the dialogue 'run in terminal' will execute as POSIX dash.

The benefit of specifying the `/usr/bin/env sh` at the beginning tells the system to use `sh` instead of whatever shell you yourself are actively using at the time of script execution.

[...]

PR #98, opened 2020-01-07
Use . instead of source in build-push.sh

gucharbon:
This is the pull request related to issue #97


tiangolo:
Yep, good point. Thanks for your contribution! :rocket: :tada:

PR #102, opened 2020-01-19
simplify update_user_me

scil:


{{cookiecutter.project_slug}}/backend/app/app/api/api_v1/endpoints/users.py
```diff
@@ -58,22 +58,19 @@ def create_user(
...
-    current_user_data = jsonable_encoder(current_user)
-    user_in = UserUpdate(**current_user_data)
-    if password is not None:
-        user_in.password = password
-    if full_name is not None:
-        user_in.full_name = full_name
-    if email is not None:
-        user_in.email = email
+    if user_in.password is None:
+        user_in.password = current_user.password
```
ebreton: Actually, a DBUser does not have a password (thanksfully), but an hashed password 🙈

tiangolo:
The problem is that the idea is that a user can only update his/her own `password`, `full_name` and `email`.

By setting the body to `UserUpdate` that would allow (or at least give the impression that it would allow) the user to also update `is_superuser`, `is_active`, etc.

So I'm gonna pass on this one. But thanks for your interest!

PR #107, opened 2020-02-06
Enable updating the password of a user.

pspoerri:
`UserUpdate` does not have a `hashed_password`.

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
+++ b/{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
@@ -5,7 +5,7 @@ from sqlalchemy.orm import Session
 from app.models.user import User
 from app.schemas.user import UserCreate, UserUpdate
 from app.core.security import verify_password, get_password_hash
-from app.crud.base import CRUDBase
+from app.crud.base import CRUDBase, ModelType, UpdateSchemaType
 
 
 class CRUDUser(CRUDBase[User, UserCreate, UserUpdate]):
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
```diff
@@ -24,6 +24,11 @@ def create(self, db_session: Session, *, obj_in: UserCreate) -> User:
         db_session.refresh(db_obj)
         return db_obj
 
+    def update(self, db_session: Session, *, db_obj: ModelType, obj_in: UpdateSchemaType) -> ModelType:
+        if obj_in.password:
+            db_obj.hashed_password = get_password_hash(obj_in.password)
```
pspoerri: I am not sure if this is the correct place to update the password.

ebreton: You probably want to use the `reset_password` method in [login.py](https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/api_v1/endpoints/login.py)

laurentS:
I was about to open almost the same PR to fix #109. I've put a small diff in there that you might want to add to your PR to fix the test as well :)

{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
pspoerri: > You probably want to use the `reset_password` method in [login.py](https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/api_v1/endpoints/login.py)

Shouldn't it be the other way around, e.g. the endpoint/login.py using CRUDUser?

tiangolo:
Thanks for your contribution! :rocket:

I think this was solved/included in https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/106 , right?

pspoerri:
Yes, looks like it. 😅

tiangolo:
Thanks for reporting back and closing it :+1:

Issue #115, opened 2020-02-22
I can't get the tests running


robido:
I am new with Docker so maybe its a noob issue, I have no clue on how to run the tests. Here is what I tried, with corresponding outputs:

`docker-compose exec backend-tests /tests-start.sh` 
‘bash\r’: No such file or directory

`docker-compose exec backend-tests \tests-start.sh`
OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "exec: \"\\\\tests-start.sh\": executable file not found in $PATH": unknown

`docker-compose exec backend-tests tests-start.sh`
OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "exec: \"tests-start.sh\": executable file not found in $PATH": unknown

`docker-compose exec backend bash`
`root@0dca8fc07bda:/app# DOMAIN=backend sh ./scripts/test.sh`
sh: 0: Can't open ./scripts/test.sh

`root@0dca8fc07bda:/app# sh scripts/test.sh`
sh: 0: Can't open scripts/test.sh

`root@0dca8fc07bda:/app# ls`
Pipfile  alembic  alembic.ini  app  prestart.sh  scripts  tests-start.sh  worker-start.sh

`root@0dca8fc07bda:/app# sh tests-start.sh`
tests-start.sh: 2: set: Illegal option -

slushnys:
that line should be ran from your local computer of your root directory.

Question offtopic:
I'm questioning the way this is tested, each time you want to tests, you must spawn new docker-compose? Where's productivity in that?

robido:
Thanks @slushnys can you elaborate more what you mean by root directory? Do you mean within bash within the container.

slushnys:
You have to execute DOMAIN=backend sh ./scripts/test.sh in your main folder which will spawn another backend container where it will run a script to execute pytest. After that you can enter the backend-tests container and run pytest whenever you want to 👍🏻

robido:
The command `DOMAIN=backend sh ./scripts/test.sh` does not work for me, I assume there is some difference between Linux and Windows for those commands. That is the error I get:
'DOMAIN' is not recognized as an internal or external command,
operable program or batch file.

I finally was able to run the tests. To help others with the same issue, here is what worked: 
[...]

Issue #41, opened 2019-06-16
Backend not starting


olegasdo:
Hello
I've tried running project locally on Windows server (using cookiecutter), but backend is not starting.
docker logs shows:
Checking for script in /app/prestart.sh
Running script /app/prestart.sh
: not foundad.sh: 2: /app/prestart.sh:

olegasdo:
Sorry, was using docker logs
"docker-compose logs backend" shows all the texts :)
Attaching to test_project_backend_1
backend_1        | Checking for script in /app/prestart.sh
backend_1        | Running script /app/prestart.sh
: not found      | /start-reload.sh: 2: /app/prestart.sh:

olegasdo:
there was an issue with  backend/app/prestart.sh
when executing directly in container it returned error "/usr/bin/env: ‘bash\r’: No such file or directory"
changing line endings in vscode from "CRLF" to "LF" solved the issue. Project files might need update

mastacheata:
I think that's an issue on cookiecutter and not on the files in this repository.

tiangolo:
Indeed, it's probably an issue with Cookiecutter on Windows: https://github.com/cookiecutter/cookiecutter/pull/1159

But still, this should be fixed now with https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/149 🎉 

It fixes the line endings of shell scripts right after generation.

rohitkumar93:
I changed my prestart.sh to "LF" using VScode too, and checked it as well. 
But I am getting the same error:
`Checking for script in /app/prestart.sh

Running script /app/prestart.sh

/start-reload.sh: 22: .: Can't open /app/prestart.sh`

pranjalsrv:
This is in the output of command `docker-compose logs`:

`backend_1  | Checking for script in /app/prestart.sh
backend_1  | There is no script /app/prestart.sh
backend_1  | INFO:     Uvicorn running on http://0.0.0.0:80 (Press CTRL+C to quit)
backend_1  | INFO:     Started reloader process [1] using statreload
backend_1  | ERROR:    Error loading ASGI app. Import string ":app" must be in format "<module>:<attribute>".`

This issue still persists on running on Windows. It is the duplicate of 
I tried changing the line endings using both VSCode and manually using the command:`find . -type f -exec dos2unix {} \;`, but the issue still exists.

pranjalsrv:
This is in the output of command `docker-compose logs`:

[...]

Issue #13, opened 2019-04-17
Cookiecutter Windows Bug in worker-start.sh: CRLF Linebreaks


Sieboldianus:
First, many thanks for this project! I am lookin forward to testing it.

I had some problems getting it to run on Windows and found the source (it was me). Even though it was my fault, I thought a note in the readme could be added for other users:

The backend would not start because linebreaks in `worker-start.sh` (and other bash-files) were CRLF (Windows), and not LF (Linux). In `docker-compose logs` this would appear as `invalid command \r` etc.

I used cookiecutter in Windows with `cookiecutter https://github.com/tiangolo/full-stack-fastapi-postgresql`, which issued CRLF linebreaks. Once using cookiecutter from WSL bash in Ubuntu, everything worked fine!

I think it generally a good recommendation to use docker-compose, and docker from WSL bash (connected through Windows Docker daemon running on localhost); and set VSCode (if used from Windows), to use LF in global settings with `"files.eol": "\n"`.

tiangolo:
Thanks! I think it's a problem in Cookiecutter itself, this PR seems to fix it: https://github.com/audreyr/cookiecutter/pull/1159

tiangolo:
This should be fixed by https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/149 🎉

PR #118, opened 2020-02-25
Update README.md to help Windows users

robido:
#115 #41 #13 

The extra comments in README.md will help Windows users get past the above issues, until this is fixed in Cookiecutter:
https://github.com/cookiecutter/cookiecutter/pull/1159

### DIFF
--- a/{{cookiecutter.project_slug}}/README.md
+++ b/{{cookiecutter.project_slug}}/README.md
@@ -49,6 +49,18 @@ docker-compose logs backend
 
 If your Docker is not running in `localhost` (the URLs above wouldn't work) check the sections below on **Development with Docker Toolbox** and **Development with a custom IP**.
 
+## Windows development notes
+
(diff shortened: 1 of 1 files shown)


AkashV10:
Hi Robido, 
Maybe you can add the following command to convert all line endings 
```bash
find . -type f -exec dos2unix {} \;
```

This would be easier than doing it with notepad++

source - 
https://stackoverflow.com/questions/7068179/convert-line-endings-for-whole-directory-tree-git

tiangolo:
Thanks for taking the effort to document this @robido ! 🤓 🍰 

I finally found a way to solve it in the project itself with https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/149 🎉 

Now, for new projects, it fixes the line endings of shell files after project generation.

Given that, I'll pass on this PR. But thanks for your effort!

PR #126, opened 2020-03-03
Simplify Dockerfiles with multi-stage builds

michael0liver:
Leveraging [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) in Docker, condense the three existing dockerfile's into one.

The compose files version also needed a bump to support the `target` option.

I haven't touched the pip requirements but they could be broken out to requirements files and updated too.


tiangolo:
Thanks for your interest!

I'm actually not sure I understand well what would be the benefit of this :thinking:

michael0liver:
End goal was to break the requirements out into requirements.txt for pip or pyproject.toml for poetry. Then combined with a single Dockerfile it would reduce repetition. I see you plan to remove the tests.dockerfile so maybe it will be less helpful now. 

I will close 😊

tiangolo:
Get it, thanks!

PR #127, opened 2020-03-07
add pypi mirror option

ercJuL:
if set the pip_mirror_url will be use to download, when building backend image


tiangolo:
Thanks for your interest!

I would prefer to keep this as a customization done on each individual project when necessary. I'm trying to reduce the number of config options to simplify everything, to reduce the maintenance effort and the complexity for someone new.

So I'll pass on this one, but thanks for your effort! :cake:

PR #128, opened 2020-03-07
Hook to hide credentials in gitignore

alexmitelman:
This PR should solve [this issue](https://github.com/tiangolo/full-stack-fastapi-postgresql/issues/122).

We create cookiecutter post-generate [hook](https://cookiecutter.readthedocs.io/en/1.7.0/advanced/hooks.html) that simply adds files with credentials to `.gitignore`

Tested using my fork branch:
`cookiecutter https://github.com/alexmitelman/full-stack-fastapi-postgresql.git`

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -2,7 +2,7 @@ import sys
 from pathlib import Path
 
 
-files_to_ignore = '''
+files_to_ignore_text = '''
 # Warning! Files in the below section contain your credentials!
 # Credentials should never be pushed to the repository.
 cookiecutter-config-file.yml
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for your suggestion and contribution! :cake: :rocket: 

Nevertheless, I think it depends on each use case, if you need to hide those secrets, like in a public repo, you can put them in `.gitignore`. But you have to make sure they are available in your CI for testing, in your deployment CI or infrastructure, that you share them in some way with other developers, etc.

If you use a private repo and deploy to your private cloud, then it's not a problem to have them there.

That's why I don't enforce any git behavior with those files, but you are free to add them to `.gitignore` locally if you want to.

Given that, I'm gonna pass on this. But thanks for your help and contribution here!

PR #129, opened 2020-03-10
dashboard welcome message to show email if full name doe not exists

rlonka:
In Dashboard the greeting did not show email in case UserProfile exists without fullname.
The condition never came to the else statement


tiangolo:
Ah, good catch, thank you! :detective: :mag: :rocket: :cake:

PR #135, opened 2020-04-01
Remove duplicate 'login' tags

Nonameentered:
The login tag is already included here.
https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/1d8678235d5da2911b8e279736f527644b1797ac/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/api_v1/api.py#L6


tiangolo:
Nice, thank you! :fire: :cake:

PR #149, opened 2020-04-18
Fix Windows line endings for shell scripts after generation

tiangolo:
:bug: Fix Windows line endings for shell scripts after generation.

This uses Cookiecutter hooks and pathlib to replace all CR LF characters with LF in `*.sh` files.

https://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -1,0 +1,8 @@
+from pathlib import Path
+
+
+path: Path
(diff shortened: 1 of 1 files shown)


joehoeller:
path: Path
        ^
SyntaxError: invalid syntax
ERROR: Stopping generation because post_gen_project hook script didn't exit successfully
Hook script failed (exit status: 1)

PR #159, opened 2020-04-20
Format fixes

tiangolo:
:art: Extra format fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,17 @@ After using this generator, your new project (the directory created) will contai
 
 ### Latest Changes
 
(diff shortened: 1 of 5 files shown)


PythonWongXb:
Sir,can u tell me why u write down something in init.py sometimes, but sometimes u write down nothing in init.py,I can’t understand it  sir, could u tell a little Pythonista the reason,thx.

tiangolo:
@PythonWongXb hey, you can check the docs here: https://docs.python.org/3/tutorial/modules.html#packages

PR #165, opened 2020-04-22
:recycle: Refactor frontend

cbhagl:
Hi,

thanks for the great work on FastAPI and this template. We are using it in my team at [Hella Aglaia Mobile Vision GmbH](https://hella-aglaia.com) to build an internal tool for our machine learning services.
While working with the frontend template, I have noticed that some of the dependencies are outdated, so I started refactoring the frontend to bring everything up-to-date. Here's what I did.

* :recycle: Migrate to latest Vuetify 2.x API, for details see migration guide (https://vuetifyjs.com/en/getting-started/releases-and-migrations/#migration-guide)
* :recycle: Migrate to latest Vee-Validate 3.x API, for details see migration guide (https://logaretm.github.io/vee-validate/migration.html#migration-guide)
* :heavy_plus_sign: Refactor state management with vuex-module-decorators (https://github.com/championswimmer/vuex-module-decorators), which is similar in style to vue-class-components
* :sparkles: Add configuration for eslint (+prettier, +typescript)
* :art: Reformat code with `eslint --fix`
* :pushpin: Upgrade/add relevant dependencies
* :truck: Remove external dependencies (e.g. mdi fonts) by adding them as direct dependencies

Maybe there is interest in these changes.

Cheers


Nonameentered:
I'm still learning frontend development at the moment, could you or someone else explain why the test is failing?

If we switch to eslint, do we still need [tslint.json](https://github.com/cbhagl/full-stack-fastapi-postgresql/blob/83310a6e54dcfc6d5ae7e0be1e31731bc68bc98e/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/tslint.json)?

When I do `npm run lint` I get errors like these in [shims-tsx.d.ts](https://github.com/cbhagl/full-stack-fastapi-postgresql/blob/83310a6e54dcfc6d5ae7e0be1e31731bc68bc98e/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/src/shims-tsx.d.ts#L1) although I don't have any issues when running `vue-cli-service serve`:
```
error: 'Vue' is defined but never used. Allowed unused vars must match /^_/u (@typescript-eslint/no-unused-vars) at src/shims-tsx.d.ts:1:8:
> 1 | import Vue, { VNode } from "vue";
    |        ^
  2 | 
  3 | declare global {
  4 |   namespace JSX {


error: 'VNode' is defined but never used. Allowed unused vars must match /^_/u (@typescript-eslint/no-unused-vars) at src/shims-tsx.d.ts:1:15:
> 1 | import Vue, { VNode } from "vue";
    |               ^
  2 | 
  3 | declare global {
  4 |   namespace JSX {


warning: Interface name must be prefixed with "I" (@typescript-eslint/interface-name-prefix) at src/shims-tsx.d.ts:6:15:
  4 |   namespace JSX {
  5 |     // tslint:disable no-empty-interface
> 6 |     interface Element extends VNode {}
    |               ^
  7 |     // tslint:disable no-empty-interface
  8 |     interface ElementClass extends Vue {}
  9 |     interface IntrinsicElements {
```
Should this be happening?

{{cookiecutter.project_slug}}/frontend/src/main.ts
```diff
@@ -1,19 +1,15 @@
...
-import './component-hooks';
-import Vue from 'vue';
-import './plugins/vuetify';
-import './plugins/vee-validate';
-import App from './App.vue';
-import router from './router';
-import store from '@/store';
-import './registerServiceWorker';
-import 'vuetify/dist/vuetify.min.css';
+import Vue from "vue";
```
Nonameentered, line 1: This import is necessary to get vue router redirects working again (so that  `beforeRouteEnter` gets called)
```suggestion
import "./component-hooks";
import Vue from "vue";
```

cbhagl:
Hi @Nonameentered,
 
thank you very much for pointing out these issues! I too have just started front-end development, so I really appreciate someone peer-reviewing my changes. 

> If we switch to eslint, do we still need [tslint.json](https://github.com/cbhagl/full-stack-fastapi-postgresql/blob/83310a6e54dcfc6d5ae7e0be1e31731bc68bc98e/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/tslint.json)?

You are right, `tslint.json` is obsolete if we switch to eslint. I missed that!

> When I do `npm run lint` I get errors like these in [shims-tsx.d.ts](https://github.com/cbhagl/full-stack-fastapi-postgresql/blob/83310a6e54dcfc6d5ae7e0be1e31731bc68bc98e/%7B%7Bcookiecutter.project_slug%7D%7D/frontend/src/shims-tsx.d.ts#L1) although I don't have any issues when running `npm run serve`:
>
> [...]
> 
> Should this be happening?

These shims are auto-generated by `vue-cli` and I think they can safely be ignored by eslint. I will add the corresponding configuration.

Regarding the changes you suggested, they all look valid. I will incorporate them into the pull request :)

Thanks again!

cbhagl:
> Not sure how to/if it's possible to suggest these changes since it wasn't already edited in the PR.

I think one way could be to do a Pull Request on my fork, but not sure.

Nonameentered:
Looks like the Manage Users page is broken by this update, in that you can't click to edit the user through the actions column.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
```diff
@@ -8,76 +8,74 @@
...
+      {
+        text: "Full Name",
+        sortable: true,
+        value: "full_name",
+        align: "left",
+      },
+      {
+        text: "Is Active",
+        sortable: true,
+        value: "isActive",
```
Nonameentered, line 54:
```suggestion
        value: "is_active",
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
```diff
@@ -8,76 +8,74 @@
...
+        align: "left",
+      },
+      {
+        text: "Is Superuser",
+        sortable: true,
+        value: "isSuperuser",
```
Nonameentered, line 60:
```suggestion
        value: "is_superuser",
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
```diff
@@ -8,76 +8,74 @@
...
+        align: "left",
+      },
+      {
+        text: "Actions",
+        value: "id",
```
Nonameentered, line 65:
```suggestion
        value: "actions",
```

cbhagl:
Ah yes, I've been meaning to address this issue, but hadn't found time to do so yet.

Naming conventions in JavaScript mostly require `camelCase` variables, whereas in Python we like `snake_case`. This will of course trigger a linter on either side, depending which style you use. In this case, we choose `snake_case` which will satisfy flake8 (with plugin `pep8-naming`) but trigger eslint (as configured in this pull request). 
However, there is a workaround for this, which will satisfy both: Pydantic models can be configured with an [`alias_generator`](https://pydantic-docs.helpmanual.io/usage/model_config/#alias-generator) that takes a function which converts all field names (e.g. a function that converts snake to camel case). See an example implementation of this here: https://github.com/nsidnev/fastapi-realworld-example-app/blob/master/app/models/domain/rwmodel.py
This would make all schemas compatible with PEP8 in the backend and eslint in the frontend.

For now though, your changes are correct.

monatis:
Hi @tiangolo, any plan to merge this pr? Or should we add it to our fork?

monatis:
I forked @cbhagl's `refactor-frontend` branch and then pulled @tiangolo's master and pushed it into my [fork](https://github.com/monatis/full-stack-fastapi-postgresql). Anyone who wants to start with Vuetify 2 can use this one until this PR is merged.

Biskit1943:
@Nonameentered why did you used `~` for some dependencies? Isn't `^` better, as this would pull new features without breaking anything? Or am I wrong here?

cbhagl:
I think vue-cli deliberately uses `~` for the plugins it manages.

Biskit1943:
Ok, maybe I mentioned the wrong user ^^ Anyway I got the right person, why don't we replace the `~` with `^`. I don't know JS that much but it seems to be better to use `^` for the dependencies as we could only benefit from the new minor versions.
Also this project serves as a template, therefore it would be more future proof to use `^` instead of `~`.

cbhagl:
I think vue-cli wants you to use `vue upgrade` to upgrade its plugins because those upgrades might involve some changes that need the users attention.

Biskit1943:
@cbhagl didn't know that, thanks for the clarification.

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      const loadingNotification = { content: "saving", showProgress: true };
+      mainStore.addNotification(loadingNotification);
+      const response = (
+        await Promise.all([
+          api.createUser(mainStore.token, payload),
+          await new Promise((resolve, _reject) => setTimeout(() => resolve(), 500)),
+        ])
+      )[0];
+      mainStore.setUserProfile(response.data);
```
Biskit1943, line 76: Executing this will overwrite the current Profile with the new created Profile. Therefore you're no longer a superuser (or your token is but the stored profile is not). Navigating to `/main/profile/view` will result in the wrong (new create) Profile.
Removing this line should fix this.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
```diff
@@ -1,145 +1,187 @@
...
+
+              <!-- is active -->
+              <div class="subheading secondary--text text--lighten-2">
+                User is active <span v-if="isActive">(currently active)</span
+                ><span v-else>(currently not active)</span>
+              </div>
+              <v-checkbox v-model="isActive" label="Is Active"></v-checkbox>
+              <v-row align="center">
+                <v-col class="flex-shrink-1">
+                  <v-checkbox v-model="setPassword" class="mr-2"></v-checkbox>
```
Biskit1943, line 63: A label would be nice here, so you know for what the button ist.
For Example:
```vue
                  <v-checkbox
                    v-model="setPassword"
                    class="mr-2"
                    label="Override Password"
                  ></v-checkbox>
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
```diff
@@ -1,145 +1,187 @@
...
+                    name="Password confirmation"
+                    vid="password2"
+                    :rules="{
+                      required: setPassword,
+                      confirmed: setPassword ? 'password1' : false,
+                    }"
+                  >
+                    <v-text-field
+                      v-model="password2"
+                      type="password"
```
Biskit1943, line 97: Like in `password1`, would it be nice, if this input field is also disabled if `setPassword` is false.
```vue
                    <v-text-field
                      v-model="password2"
                      type="password"
                      :disabled="!setPassword"
                      label="Confirm Password"
                      :error-messages="errors"
                    ></v-text-field>
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
```diff
@@ -8,76 +8,74 @@
...
-        <td><v-icon v-if="props.item.is_superuser">checkmark</v-icon></td>
-        <td class="justify-center layout px-0">
-          <v-tooltip top>
-            <span>Edit</span>
-            <v-btn slot="activator" flat :to="{name: 'main-admin-users-edit', params: {id: props.item.id}}">
-              <v-icon>edit</v-icon>
-            </v-btn>
-          </v-tooltip>
-        </td>
+      <template v-slot:item.is_active="{ item }">
```
Biskit1943, line 11: My editor (VS-Code) gives me the following error: `'v-slot' directive doesn't support any modifier`. I googled this and found the following answer: https://stackoverflow.com/a/63309643
Which solved this "issue" for me.

Except of this project I never programmed in JS, therefore I don't know the exact reason and if this solution is appropriate. However, I changed my code to look like this (I hate red warnings 😛):
```vue
    <v-data-table :headers="headers" :items="users">
      <template v-slot:[`item.is_active`]="{ item }">
        <v-icon v-if="item.is_active">mdi-check</v-icon>
      </template>

      <template v-slot:[`item.is_superuser`]="{ item }">
        <v-icon v-if="item.is_superuser">mdi-check</v-icon>
      </template>

      <template v-slot:[`item.actions`]="{ item }">
        <v-icon @click="routeEditUser(item.id)">mdi-pencil</v-icon>
      </template>
```

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
```diff
@@ -0,0 +1,252 @@
...
+  async checkApiError(payload: AxiosError) {
+    if (payload.response && payload.response.status === UNAUTHORIZED) {
+      await this.logOut();
+    }
+  }
+
+  @Action
+  async routeLoggedIn() {
+    if (router.currentRoute.path === "/login" || router.currentRoute.path === "/") {
+      router.push("/main");
```
Biskit1943, line 188: This leads to an error (even though everything works fine):
```
Uncaught (in promise) Error: Redirected when going from "/login" to "/main" via a navigation guard.
```

Changing this to `/main/dashboard` would remove this error but would also make it more work to change the "main" page, as this line has to change too.

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
```diff
@@ -0,0 +1,252 @@
...
+import router from "@/router";
+import { AxiosError } from "axios";
+import { VuexModule, Module, Mutation, Action } from "vuex-module-decorators";
+import { IUserProfile, IUserProfileUpdate, IAppNotification } from "@/interfaces";
+import { UNAUTHORIZED } from "http-status-codes";
+
+@Module({ name: "main" })
+export default class MainModule extends VuexModule {
+  token = "";
+  isLoggedIn: boolean | null = null;
```
Biskit1943, line 12: I don't know why isLoggedIn should ever be null? Isn't is either false or true (false by default)?

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
```diff
@@ -0,0 +1,252 @@
...
+  logInError = false;
+  userProfile: IUserProfile | null = null;
+  dashboardMiniDrawer = false;
+  dashboardShowDrawer = true;
+  notifications: IAppNotification[] = [];
+
+  get hasAdminAccess() {
+    return (
+      this.userProfile && this.userProfile.is_superuser && this.userProfile.is_active
```
Biskit1943, line 21: Adding `this.isLoggedIn && ` to this function will fix a Bug where you log in as admin and then logout. The drawer will now still hold the `Admin` items (I disabled the mandatory login). This but could also be fixed in `removeLogIn()`, see below.

These changes wouldn't hurt any developer keeping the login guard but would definitely help those removing it.

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
```diff
@@ -0,0 +1,252 @@
...
+        await this.removeLogIn();
+      }
+    }
+  }
+
+  @Action
+  async removeLogIn() {
+    removeLocalToken();
+    this.setToken("");
+    this.setLoggedIn(false);
```
Biskit1943, line 156: Adding `this.setUserProfile(null)` would fix the Bug mentioned above. Consider setting `setUserProfile(palyload: IUserProfile | null)` so no warning is printed.

Biskit1943 (changes requested):

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
cbhagl: It would be nice if you could make a change suggestion like @Nonameentered did previously. Then I can accept and apply the change right away.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
cbhagl: I think this is a change that would generally be beneficial to the main branch as well. I don't want to merge further improvements into this PR, which might not end up in main after all.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
cbhagl: What version of eslint are you using?
VScode probably complains because either the language server is unhappy or a linter emits a warning. In this case I think it is the latter (eslint) and the warning you are describing must be a new default, because I did not see this before.
From my understanding of Vue this is indeed [valid syntax](https://vuejs.org/v2/guide/components-slots.html#Dynamic-Slot-Names) and would fix the warning. However, I think it is kind of ugly in this case and maybe suppressing the warning (either via inline comment or global setting) would be better?

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
cbhagl: This is also something I think could be addressed in the main branch as a general improvement.

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
cbhagl: This is also something taken over from the main branch. `null` is the default state for `isLoggedIn` and you have an explicit `isLoggedIn===null`-check in `App.vue` which determines if the circular progress bar should be shown or not.
I think this is used to indicate the initial state (ie. the app is still trying to connect to the authentication backend).

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
cbhagl: Again, something for the main branch.

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
cbhagl: I would like to incorporate this change. Please make a change suggestion so I can accept it. While you're at it, you can add a little syntactic sugar by using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
Biskit1943: I will do so. This was my first real review on github and I did not know that this button exists 🙈.
```suggestion
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
Biskit1943: I disagree with that. Even if this would be changed on the master, merging your pull request would remove this again. Therefore I'd like to include this right here and maybe add extra pull request to the master for this.

If you want to, you can accept this here or wait for the merge and change it later:
```suggestion
                  <v-checkbox
                    v-model="setPassword"
                    class="mr-2"
                    label="Override Password"
                  ></v-checkbox>
```

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
Biskit1943: As with the other changes, I think including this would be good and I will also address this in a extra PR.

```suggestion
      router.push("/main/dashboard");
```

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
Biskit1943: Ok, this seems correct then :)

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
Biskit1943:
```suggestion
      this.isLoggedIn && this.userProfile?.is_superuser?.is_active
```
I'm not quite sure, if the third access to `is_active` works (can't test right now)

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
Biskit1943:
```suggestion
    this.setLoggedIn(false);
    this.setUserProfile(null);
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
Biskit1943: My version is `v6.8.0` and indeed, the "correct" syntax isn't that beautiful. I'll provide a request change with the corresponding ignore comment.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/EditUser.vue
cbhagl: Even if that is the case, it might be worth it to do a separate PR which I could merge into my branch as well. So if either gets accepted, The changes will be included. It might be redundant, but this way we can make sure everybody will benefit from these improvements eventually.

Nonetheless, I will accept these now, but would be happy if you write a separate PR as well. Small changes like this are easily reviewed and probably quick to merge.

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
cbhagl: > this.isLoggedIn && this.userProfile?.is_superuser?.is_active

Yes, this shouldn't work as it would mean `is_superuser` is an object or `undefined`, containing a `is_active` key, which is not the case.

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
cbhagl: Seems like others run into this problem as well: https://github.com/vuejs/eslint-plugin-vue/issues/260#issuecomment-668794653

cbhagl: Have you tried `<!-- eslint-disable-next-line eslint-plugin-vue/valid-v-slot -->`

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
```diff
@@ -0,0 +1,85 @@
...
+      await mainStore.checkApiError(error);
+    }
+  }
+
+  @Action
+  async createUser(payload: IUserProfileCreate) {
+    try {
+      const loadingNotification = { content: "saving", showProgress: true };
+      mainStore.addNotification(loadingNotification);
+      const response = (
```
cbhagl, line 70:
```suggestion
      const _response = (
```

cbhagl (commented):
Build fails because `response` is now unused.

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
```diff
@@ -0,0 +1,85 @@
...
+        await Promise.all([
+          api.createUser(mainStore.token, payload),
+          await new Promise((resolve, _reject) => setTimeout(() => resolve(), 500)),
+        ])
+      )[0];
```
cbhagl: Remove declaration of unused `response` entirely.

```suggestion
      await Promise.all([
        api.createUser(mainStore.token, payload),
        await new Promise((resolve, _reject) => setTimeout(() => resolve(), 500)),
      ]);
```

{{cookiecutter.project_slug}}/frontend/src/store/modules/main.ts
```diff
@@ -0,0 +1,253 @@
...
+    this.isLoggedIn = payload;
+  }
+
+  @Mutation
+  setLogInError(payload: boolean) {
+    this.logInError = payload;
+  }
+
+  @Mutation
+  setUserProfile(payload: IUserProfile) {
```
cbhagl, line 45: Fix errorwhen trying to set user profile to `null`.

```suggestion
  setUserProfile(payload: IUserProfile | null) {
```

cbhagl (commented):
Fix build errors.

{{cookiecutter.project_slug}}/frontend/src/store/modules/admin.ts
cbhagl: Fix build errors (again).
```suggestion
      await Promise.all([
        api.createUser(mainStore.token, payload),
        await new Promise((resolve, _reject) => setTimeout(() => resolve(), 500)),
      ]);
```

{{cookiecutter.project_slug}}/frontend/src/views/main/admin/AdminUsers.vue
Biskit1943: Ok, I checked if a update of everything to the latest version fixes this and it did not. *But* I also opened the project in PyCharm Professional and the error was not linted at all. So it seems to me, that this is a problem of VS-Code? Anyway, as this problem is not present in all editors I wouldn't change it.

cbhagl: Try running `eslint` directly on the file (via command-line) and see if the error appears. Correctly configured editors will generally output the same warnings as they are just displaying what either the vue language server or (in our case) eslint say. I think the error you are seeing is some new default rule (ie. vue/valid-v-slot) that eslint-plugin-vue complains about and there is an bug in the plugin that prohibits correctly disabling it.

wert1cmd:
I refactored my project by hand using your code here as a guide and it worked splendidly.

Excellent work! Thank you.

tiangolo:
Thanks for all the work here @cbhagl! :bow: 

I wanted to refactor the frontend and move to React with TypeScript and hooks, etc. It's now been done in other PRs, so I'll pass on this one. But thanks for all the effort! :cake:

PR #168, opened 2020-04-26
Fixes #166 update user fails when no password

### LINKED ISSUE #166: Update user fails when not provided a password
Logged into the vue UI and attempted to change the Full Name value resulted in a key error for the following line

Error Output
```
backend_1       |   File "./app/crud/crud_user.py", line 34, in update
backend_1       |     if update_data["password"]:
backend_1       | KeyError: 'password'
```

https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/crud/crud_user.py#L34

The frontend sends the following as the payload

```{full_name: "Chris Allnutt", email: "chris@some_domain_here.com"}```

I'd be happy to open a PR if someone could provide direction on the fix, as far as I can tell exclude_unset combined with the check of the password is what's causing the issue.

I also think it makes sense to get rid of the `Union[UserUpdate, Dict[str, Any]` and just use UserUpdate for consistency.

My fix would be to change the check to be `if  "password" in update_data.keys():` but I'm open to other ideas as well.

cakebread:
Seems to have been introduced in commit eed33d2 during refactoring.

### DIFF
--- a/src/backend/app/api/api_v1/endpoints/users.py
+++ b/src/backend/app/api/api_v1/endpoints/users.py
@@ -1,7 +1,7 @@
 from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException
-from sqlmodel import func, select
+from sqlmodel import func, select, delete
 
 from app import crud
 from app.api.deps import (
@@ -21,6 +21,7 @@ from app.models import (
     UsersOut,
     UserUpdate,
     UserUpdateMe,
+    Item
 )
 from app.utils import send_new_account_email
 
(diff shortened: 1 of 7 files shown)


davidolrik:
How about simply `if update_data.get("password", ""):` ?

VictorBac:
I had this exact problem, fixed it with :
`if update_data.get("password"):`

Hultner:
> How about simply `if update_data.get("password", ""):` ?

I use this but `if update_data.get("password", default=False):` to be even more explicit and avoid implicit type conversion.

PR #172, opened 2020-04-29
issue#171 decoded jwt token returns 'email' which is not present

mehedees:
now returned 'sub'. probably caused by 6fdba196397d9896bce0fee80bbb15c479d40e45


tiangolo:
Thanks @mehedees! This was handled in another PR, so I'll close this one, but thanks for the effort! :cake:

PR #179, opened 2020-05-10
extending doc to turn off migrations Closes #178

### LINKED ISSUE #178: Doc on turning migrations off
Documentation fix:

Uncommenting the line below does not do the trick to turn off migrations as described in the documentation.

```python 
Base.metadata.create_all(bind=engine)
```

factorlive:
The commented line requires reference to a previously refactored base.py and an additional import to refer to the database engine.

The Readme documents that alembic migration can be turned off and hence this extension of the documentation and commented code allows to do so going forward.

:dancers: When working together in a repo, do consider to use migrations to avoid database schema conflicts that cannot be rolled back.

### DIFF
--- a/{{cookiecutter.project_slug}}/README.md
+++ b/{{cookiecutter.project_slug}}/README.md
@@ -259,10 +259,12 @@ $ alembic revision --autogenerate -m "Add column last_name to User model"
 $ alembic upgrade head
 ```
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks! This was covered in another PR, so I'll pass on this one, but thanks for the effort! :coffee:

PR #183, opened 2020-05-11
Make UserInDBBase require id

Nonameentered:
Since this is the primary key for the user table in the db, it should be safe to make this a required item here (like in schema/items.py).


tiangolo:
Thanks @Nonameentered! This was all refactored in another PR to use SQLModel, and this is handled in the new code. Thanks for the work! :cake:

PR #184, opened 2020-05-11
db.refresh() is unnecessary after commits

Nonameentered:
When we do db.commit(), this should automatically refresh the db session so there's no need to db.refresh again.


Nonameentered:
Hmm, since the tests for this fail on crud_update functions, perhaps my understanding of how sessions work is flawed. I'll close this for now until I understand why this isn't working here, even though tests work on my project without using db.refresh() in most functions.

tiangolo:
Thanks for reporting back and closing the issue!

PR #185, opened 2020-05-11
Return items in user calls

Nonameentered:
This shows an example of how to leverage relationships already created in the sqlalchemy models to return a user's items in responses that return User.


TheCDC:
Seems convenient but could easily become unwieldy as a user's collection of items grows into the thousands. At that point it might make more sense to continue using the paginated Items endpoint.

tiangolo:
Thank you! This was recently heavily refactored, I plan on something similar but that will be in future updates. Given that, I'll pass on this one, but thanks for the interest! :coffee:

PR #191, opened 2020-05-17
Single dockerfile for celery worker and backend

gaganpreet:
I removed the celeryworker Dockerfile and now it uses the same image as backend.

The celery worker Dockerfile has the same steps as the backend Dockerfile, the only difference is that it's run using a different command. But with two dockerfiles steps like downloading and installing packages happen twice (when running `docker-compose build --parallel`), each for backend and celery worker. We don't need that. :-)


tiangolo:
Thank you! :nerd_face: 

The project didn't have a real use for Celery, and I currently wouldn't really recommend Celery, I just removed it completely. Probably later once I figure what would be the best alternative for use cases that need it I will update it here in the repo.

For now, as I'm removing Celery completely, I'll pass on this PR, but thanks for the effort! :cake:

PR #193, opened 2020-05-19
change `--exclude=__init__.py` to `--ignore-init-module-imports` in autoflake

NKNaaS:
If you use ` --exclude=__init__.py` to avoid removing unused imports, `__init__.py` will not be linted. Use `--ignore-init-module-imports` option.


tiangolo:
Thanks! I see you removed the repo so I suspect this is no longer relevant?

Anyway, I need to do a big overhaul in this repo :sweat_smile: 

PR #223, opened 2020-06-26
Run each test in there own transaction

Sweenu:
This is the first thing I did when I started using the template. It allows me to never have anything really committed to database when running my tests. I don't know how you run your tests but running them each in their own transaction that is then rollbacked and leave no trace whatsoever is to me the cleanest solution when running tests against a database.

I also like the fact that it gives an example of using dependency override.


tiangolo:
Thanks for your interest! Yeah, that could be interesting, it probably needs some other changes :thinking:

PR #226, opened 2020-07-05
✏️.  Set volume /var/run/docker.sock as read-only

paxcodes:
This is consistent with the `traefik.yml` in [dockerswarm.rocks](https://dockerswarm.rocks/traefik/#create-the-docker-compose-file)


paxcodes:
I thought the the proxy container refers to the traefik container in dockerswarm.rocks but according to comments on [this issue](https://github.com/tiangolo/full-stack-fastapi-postgresql/issues/116), they are meant to be 2 separate containers during deployment.

PR #236, opened 2020-07-16
#Issue 207: Can not create inactive user fix

### LINKED ISSUE #207: Can not create inactive user.
First of all, thank you very much for sharing this project. I already learned a lot by building my first API based on this template. 

When I looked at the test cases I was wondering about the intention of the following test. I thought it is about creating an inactive user but the UserCreate schema does not accept an attribute ```disabled```. The respective attribute is ```is_active```. The logic also does not make sense to me except this test intention is to show that you can not create an inactive user. Otherwise, it should be ```assert not is_active```. But then the test fails even if you replace the ```disabled``` with ```is_active```

https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/45fdd880ce8a6fff70ff057d7e0fa4918a35cb06/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/tests/crud/test_user.py#L45-L51

It turns out that CRUDUser.create ignores ```is_active```.

https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/45fdd880ce8a6fff70ff057d7e0fa4918a35cb06/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/crud/crud_user.py#L15-L21 

If it is intended that you can not create an inactive user than the User Schemas should be altered such that the ```is_active``` attribute is not optional for UserCreate.

sindwerra:
Fix #207 
1. change test_check_if_user_is_active_inactive test function, set correct attribute name, fix wrong assertion
2. update CRUDUser.create, is_active value always referred from obj_in.is_active

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
+++ b/{{cookiecutter.project_slug}}/backend/app/app/crud/crud_user.py
@@ -18,6 +18,7 @@ class CRUDUser(CRUDBase[User, UserCreate, UserUpdate]):
             hashed_password=get_password_hash(obj_in.password),
             full_name=obj_in.full_name,
             is_superuser=obj_in.is_superuser,
+            is_active=obj_in.is_active,
         )
         db.add(db_obj)
         db.commit()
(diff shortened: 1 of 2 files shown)


tiangolo:
I see you closed the PR, so I imagine this is no longer relevant, right? In any case, thanks for coming back to close it. 🤓 ☕ 

PR #251, opened 2020-08-20
Better http error codes for the login.py endpoint

machadoug:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401


codespearhead:
LGTM

julioriffel (approved):

tiangolo:
Thanks! And sorry for the delay reviewing this. 😅

There are some conflicts because this was all refactored, would you be up to updating it?

PR #258, opened 2020-09-09
Resolves #257

### LINKED ISSUE #257: Ability to run backend separately
For now its up to 2 days while I'm struggling with postgres docker container and it would've been great if there was any guidelines/ ready compose file to launch them locally and debug that locally.

orihomie:


orihomie:
Need to update stack.yml sample

tiangolo:
Thanks! This was refactored recently to simplify the Docker Compose files, so it should be much more straightforward now. Given that, I'll pass on this one, but thanks for the effort! :cake:

PR #268, opened 2020-09-15
Fix checking length of None object

orihomie:


Biskit1943:
You can't check the length of a `None` object:

```python
>>> v = None
>>> len(v)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()
```

The check for `v` is needed in order to check the length of `v`

This pull request can be closed.

kchojn (commented):
Simplified:
```   
@validator("SENTRY_DSN", pre=True)
    def sentry_dsn_can_be_blank(cls, v: str) -> Optional[str]:
        return v if v else None
```

tiangolo:
Thanks! This was updated in another PR recently, so I'll close this one now. :nerd_face:

PR #278, opened 2020-09-21
Fix KeyError in crud_user.py and add test

michaeltoohig:


Hultner:
I've had this problem as well and fixed it in pretty much the same way, reported here #285

twcurrie:
Similar: https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/245

ewae121:
Same for me

tiangolo:
Thanks! This was handled in another PR, so I'll pass on this one. But thanks for the effort! :rocket:

PR #291, opened 2020-10-09
use python3.8 for backend image

mariuz:
use python3.8 for backend image


mariuz:
Seems better https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/202

PR #323, opened 2020-11-26
Update crud_user.py

scientes:
Fixes Bug when which crashes the update function, when updating without resending the password. the current approach either the whole function errors when the dict contains no password or it suceeds when there is a password in the dict. I fixed so that it also suceeds when no Password is present. Which enables the edit user profile Page in the frontend to work again .


scientes:
duplicate

tiangolo:
Thanks for reporting back and closing the issue :+1:

PR #325, opened 2020-12-04
Fix password reset tokens

d-swift:
Use the same dictionary key for encoding/decoding password reset tokens. Currently email is encoded into the token with key `sub`, whereas decoding is attempted with `email`.


hadialqattan (approved):

twcurrie:
Duplicated: https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/246

tiangolo:
Thanks! This was handled in another PR, so I'll now pass on this one, but thanks for the interest! :coffee:

Issue #329, opened 2020-12-11
Change postgresql default port


Xevimacia:
We are working with this template for a while. And we are now bringing it into production.
We are deploying our code based on this template in Alibaba Cloud so we will be using the ApsaraDB RDS for PostgreSQL. For security reasons, we would like to change the default port from 5432 to another port. We have tried to set the environment variable `$POSTGRES_PORT` but the variable is ignored.
Can anyone point us how to customize the postgres port?

roura356a:
We are having the same issue here. We need to be able to change the default port. It would be nice to implement an environment variable like `POSTGRES_PORT` or something like that as mentioned above.

tiangolo:
Thanks! This was solved here: https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/333

PR #333, opened 2020-12-22
✨ Add support for setting `POSTGRES_PORT`

uepoch:
In some unusual setups (cloud managed databases, yay!) the default port is not used for the postgres database.

This fixes a bug in the alembic template, to properly include the content of POSTGRES_PORT in the PostgresDSN


roura356a:
We need this as mentioned in #329

tiangolo:
Thank you! :rocket: 

I updated it to align it with the rest of the current code and it's now ready. :nerd_face:

PR #343, opened 2021-02-03
changed remove id to type Any in crud base.py

hakonjs:
changed remove id to type Any


tiangolo:
Thanks @hakonjs! This was refactored in other PRs and now these classes are not used. So I'll pass on this one, but thanks for the effort! :cake:

PR #346, opened 2021-02-12
Update crud_user.py

kadimon:
fixed a bug when updating the user without updating the password


thomas-chauvet:
Hello!

Could we merge it as it doesn't dramatically change the code and is helpful to edit a user without editing password in the same time?

Thanks!

jordan-cote:
> Could we merge it as it doesn't dramatically change the code and is helpful to edit a user without editing password in the same time?
> Thanks!

@thomas-chauvet 
Maybe this hasn't been merged because of it conflicting with #348 
image

tiangolo:
Thanks! This was handled in another PR, so I'll close this one. But thanks for the effort! :coffee:

PR #365, opened 2021-04-19
🔧 Add `SMTP_SSL` option for older SMTP servers

Metrea:
Updated to cover SSL requirements for some email providers.


julioriffel (approved):

tiangolo:
Thanks! I updated it to make it TLS or SSL exclusive, as it seems both enabled would collide. :cake:

PR #366, opened 2021-04-21
Catch user update integrity error

robertop23:
Added exc.IntegrityError to catch when an user update contains an already registered email


julioriffel (approved):

tiangolo:
Thanks! Good catch! :mag: This was handled here: https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/689, preemptively checking if the new email already exists. So I'll now close this one. But thanks for the effort! :cake:

PR #367, opened 2021-04-21
Fix parsing json backend cors

robertop23:
I usually use the Python shell when I'm developing, but I cannot simple open it using the env vars because I get this error:

    pydantic.env_settings.SettingsError: error parsing JSON for "BACKEND_CORS_ORIGINS"

How to reproduce this error:

    cd backend/app/
    export $(grep -v '^#' ../../.env | tr -d ' ' | xargs) && poetry run python
> And inside the Python terminal:

    from app.core.config import settings


tiangolo:
Thanks! This should be covered by recent PRs, so I'll pass on this one, but thanks for the effort! :rocket:

PR #390, opened 2021-06-29
[FIX] Avoid unexisting password backend error

ewae121:
Issue while trying to use edit profile button.

Please let me know if I do a mistake


tiangolo:
Thanks @ewae121! :nerd_face: 

This was solved in another PR, so I'll now close this one, but thanks for the effort! :cake:

PR #394, opened 2021-07-07
Improved cookiecutter README template for newbs

timblaktu:
As a relative newbie to postgresql, sqlalchemy, and alembic, it wasn't clear to me why, when I followed the instructions for modifying models, schemas, api endpoints, and crud utils, I was still getting runtime errors when invoking the new REST endpoints I had added. Here I've made it more explicit that when making such changes to the generated project, you must create a new db revision using alembic.

Amazing work on FastAPI and this full-stack project, by the way! I have used it to go from zero to sophisticated full stack web app in about an hour. Very impressive work.


tiangolo:
Thank you!

This was refactored recently to make things much simpler. The way I had the code before was overly complex, it should be easier to understand and extend now. :nerd_face: 

As it's all a bit different now, I'll pass on this one, but thanks for the effort! :cake:

PR #438, opened 2021-11-29
⬆ Add `uvicorn[standard] to include `watchgod` and `uvloop`

alonme:
Add the "standard" extras for uvicorn, which adds the much needed watchgod for file watching.

Without it, using `--reload` uses significant CPU (almost 100% on my machine)


tiangolo:
Great, thanks! :rocket:

PR #440, opened 2021-12-13
Update crud_user.py

pycub:
Due to the fact that update() will receive its data through UserUpdate, its data may or may not consist "password". I added .get("password") to prevent any KeyError exception.


hosseinsokhandan:
password is optional in UserUpdate. It should be considered to use .get() instead of directly accessing in within crud.update()

tiangolo:
Thanks! This was handled in another PR, so I'll close this one. But thanks for the effort! :coffee:

PR #446, opened 2021-12-27
Fix Build and run issues: Polyfill deprecation and flower image start failure

kevanpeters:
From the official polypill documentation https://babeljs.io/docs/en/babel-polyfill. This will fail out at build time
```
import "core-js/stable";
import "regenerator-runtime/runtime";
```

Also fix failing flower run in docker compose up with version pin

### DIFF
--- a/{{cookiecutter.project_slug}}/docker-compose.yml
+++ b/{{cookiecutter.project_slug}}/docker-compose.yml
@@ -112,7 +112,7 @@ services:
     # You also have to change the flower command
   
   flower:
-    image: mher/flower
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks! The fix for Celery was included here https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/396

And for the frontend, it would be best in an independent PR, nevertheless, we are in the process of rebuilding the frontend so most of this code will go away soon. 🔥 

Given that, I'll close this one, but thanks for the effort! ☕

PR #453, opened 2022-02-07
fix bug on update data in curd orm

longniao:
Here is a bug for updating data in curd base.


tiangolo:
Thanks @longniao! All the CRUD logic in classes was recently refactored, so this is all now updated. Given that, I'll now close this one, but thanks for the interest! :coffee:

PR #454, opened 2022-02-13
✏ Fix typo in `{{cookiecutter.project_slug}}/backend/app/app/schemas/item.py`

ramushetty:
Properties properties  twice


tiangolo:
Thanks! This was handled in another PR, so I'll now close this one. But thanks for the effort! :cake:

PR #413, opened 2021-08-25
fix exception on user update when password is not set

Haikson:
```
backend_1       |   File "./app/crud/crud_user.py", line 34, in update
backend_1       |     if update_data["password"]:
backend_1       | KeyError: 'password'
```

PR #461, opened 2022-04-25
Update crud_user.py

nicozhang:
When I change the full name, it will show KeyError


osouza-de:
Nice solution, but unfortunately it's duplicated.

#413

tiangolo:
Thanks! This was handled in a recent PR, so I'll close this one now. But thanks for the effort! :coffee:

PR #477, opened 2022-09-16
Update LICENSE

donBarbos:


tiangolo:
Thanks, but this is not really necessary, so I'll close this one.

Issue #479, opened 2022-09-23
celeryworker.dockerfile have to change.


po2son:
celeryworker.dockerfile have to change.

```
# Install Poetry - before
# RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_HOME=/opt/poetry python && \
#    cd /usr/local/bin && \
#    ln -s /opt/poetry/bin/poetry && \
#    poetry config virtualenvs.create false

```
↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓

```
# Install Poetry - changed
RUN curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/poetry python3
ENV PATH="/opt/poetry/bin:$PATH"
RUN poetry config virtualenvs.create false
```

rabarbra:
Same with backend.dockerfile

willingc:
See PRs https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/524 and https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/529 for the corrected files if you want to take a look.

tiangolo:
Thanks! This was fixed in https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/480 🤓 

And thanks a lot for the help @willingc 🙇

PR #480, opened 2022-10-08
Fix the poetry installation in Dockerfile.

little7Li:
The poetry installation url has to be updated as the old one is no longer valid.
image

And replace a poetry install option `--no-dev` with a recommended one `--only main`.
image

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/backend.dockerfile
+++ b/{{cookiecutter.project_slug}}/backend/backend.dockerfile
@@ -3,7 +3,7 @@ FROM tiangolo/uvicorn-gunicorn-fastapi:python3.7
 WORKDIR /app/
 
 # Install Poetry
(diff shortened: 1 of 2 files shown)


little7Li:
#479

hcristea:
@little7Li, thanks for the PR, I wanted to do it myself. I'd change the title to something like "Upgrade poetry installer to v1.2" to reflect more accurately the changes. The previous installer for poetry 1.1 works fine, no need to fix it, just make an upgrade.

kwantopia:
Would be great if this were merged soon.

Luan-vP:
Bump! 

The provided URL now breaks build with a 404.

tiangolo:
Awesome, thank you @little7Li! 😄 🍰 

I updated the Python version as well as it was not building the Docker image, and I updated a couple of other dependencies needed for it to build and run. 🤓 

Thanks @hcristea for the comment! ☕

PR #512, opened 2023-05-05
Fixes to enable building

elbamos:
Fixes to enable building
- New path to download the poetry installer
- Upgrade npm to 6.9.0 to support aliases used by current version of vue


jvitali (approved):
Worked for me. I was able to run `docker compse up -d` with this changes and the temaplate went up

tiangolo:
Thank you @elbamos!

This was handled in another PR, so I'll now close this one. But thanks for the effort! :cake:

PR #524, opened 2023-07-10
replace get-poerty.py script with 'install.python-poetry.org'

snjypl:
The get-poetry.py installer has been deprecated and removed.  
need to update to : 
`curl -sSL https://install.python-poetry.org | python3 -`

refer: 
https://python-poetry.org/docs/#installing-with-the-official-installer

### DIFF
--- a/{{cookiecutter.project_slug}}/backend/backend.dockerfile
+++ b/{{cookiecutter.project_slug}}/backend/backend.dockerfile
@@ -3,7 +3,7 @@ FROM tiangolo/uvicorn-gunicorn-fastapi:python3.7
 WORKDIR /app/
 
 # Install Poetry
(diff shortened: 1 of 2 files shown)


JimFawkes:
Hello @snjypl,
just wanted to make the same change, but you beat me to it 🚀.

I think that the same change would also make sense to be done here: https://github.com/snjypl/full-stack-fastapi-postgresql/blob/patch-1/%7B%7Bcookiecutter.project_slug%7D%7D/backend/celeryworker.dockerfile#L6

tomashm (approved):

malekhnovich:
yea just ran into this too

4c0n:
@tiangolo Considering this is broken, this PR fixes it, and it has been reviewed and approved, could this please be merged?

tiangolo:
Thanks! This was solved in https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/480 🎉 

Given that, I'll close this one, but thanks for the effort. ☕

PR #525, opened 2023-07-12
orm_mode should be from_attributes

onurkanbakirci:
`orm_mode` should be `from_attributes` because it's gives 

> UserWarning: Valid config keys have changed in V2: * 'orm_mode' has been renamed to 'from_attributes'

warning.


tiangolo:
Thanks! This was probably that you installed Pydantic v2 before it was supported. But this was all recently refactored to use Pydantic v2 and SQLModel. :tada: 

Given that, I'll pass on this one, but thanks for the interest! :coffee:

PR #538, opened 2023-09-10
BugFix: api.deps.get_current_active_superuser()

nsxsnx:
The function did not check that user was active


tiangolo:
Thank you! This was recently refactored and updated and should be now handled. So I'll pass on this one, but thanks for the effort! :cake:

PR #562, opened 2023-11-29
🐛 Fix positional argument bug in `init_db.py`

alejsdev:
Fix positional argument bug in init_db.py


tiangolo:
Nice, thank you! 🍰 :tiramisu:

PR #565, opened 2023-12-05
fix(core): update config.py

maximyurevich:
Check if `SENTRY_DSN` is not `None` in validator


tiangolo:
Thanks! This was handled in another PR, so I'll now pass on this one. But thanks for the effort! :cake:

PR #572, opened 2023-12-27
♻️ Remove type ignores and add `response_model`

alejsdev:
Remove type ignores and add response_model


tiangolo:
Great, thank you! 🚀 💃

PR #574, opened 2023-12-27
♻️ Re-enable user update path operations for frontend client generation

alejsdev:
Re-enable user update path operations for frontend client generation


tiangolo:
Great, thank you! 🚀 🍪

PR #576, opened 2023-12-30
fix poetry download path in docker file, python version upgrade to 3.10 and update to the config.py

Akachi-Anabanti:
The use of lru_cache to cahe the settings this helps to improve performance of accessing app variables
The poetry download path in the backend.docker file is obsolete and therefore was updated
The python version was also updated to 3.10 to be compatible with the new poetry version
The config.py file was updated to reflect the updated dependencies for the project.


dtleal (approved):

Akachi-Anabanti:
Thank you Sebastián for your response and thank you more for this repo.

PR #577, opened 2024-01-01
Refactor update_user_me

salahelfarissi:
I refactored the update_user_me path operation.
Must use SQLModel 0.0.14.


tiangolo:
Thanks! This was handled in another PR, it required a new feature in SQLModel, but it's now there. :tada: 

Given that I'll pass on this one, but thanks for the effort! :cake:

PR #596, opened 2024-02-14
Update utils.py to fix jwt decoding during password request

nishaalnaseer:
Another way to solve this would be to edit 96 of the same file
        {"exp": exp, "nbf": now, "sub": email}, settings.SECRET_KEY, algorithm="HS256",

but this commit adheres to the convention of jwt encoding and decoding during oauth in this project


tiangolo:
Thanks! This was handled in another PR, so I'll now close this one. But thanks for the effort! :cake:

PR #597, opened 2024-02-14
Update utils.py

nishaalnaseer:
excepts key error when decoding jwt token, if a token has been tempered with and the key is missing, the server will not throw  a 500


tiangolo:
Thanks! This was recently refactored to return the `sub` required key. So I'll pass on this one, but thanks for the interest! :coffee:

PR #599, opened 2024-02-15
♻ Refactor items and services endpoints to return count and data, and add CI tests

estebanx64:
Refactor items and services endpoints and added support to run pytest within docker-compose and actions


src/backend/app/app/api/api_v1/endpoints/users.py
```diff
@@ -27,15 +28,20 @@
...
     """
+
+    statment = select(func.count()).select_from(User)
+    count = session.exec(statment).one()
+
     statement = select(User).offset(skip).limit(limit)
     users = session.exec(statement).all()
-    return users
+
+    return UsersOut(data=users, count=count)
```
andriiko489, line 44: why?

estebanx64: Front end needs to know how many items are in database to be able to create a pagination

tiangolo:
Great, thanks! I updated it a bit, mainly to match the previous GitLab CI config, mainly to avoid race conditions between migrations running and the tests starting before the app/DB is ready. This is ready! :rocket:

PR #614, opened 2024-02-27
✨ Support delete own account and other tweaks

alejsdev:
✨ Support delete own account and other tweaks


tiangolo:
Awesome, thanks @estebanx64 and @alejsdev! :rocket:

PR #615, opened 2024-02-28
⬆ Upgrade code to support pydantic V2

estebanx64:


src/.env
```diff
@@ -7,7 +7,7 @@ PROJECT_NAME="FastAPI Project"
 STACK_NAME=fastapi-project
 
 # Backend
-BACKEND_CORS_ORIGINS="[\"http://localhost\", \"http://localhost:4200\", \"http://localhost:3000\", \"http://localhost:8080\", \"https://localhost\", \"https://localhost:4200\", \"https://localhost:3000\", \"https://localhost:8080\", \"http://local.dockertoolbox.tiangolo.com\", \"http://localhost.tiangolo.com\"]"
+BACKEND_CORS_ORIGINS="["http://localhost", "http://localhost:4200", "http://localhost:3000", "http://localhost:8080", "https://localhost", "https://localhost:4200", "https://localhost:3000", "https://localhost:8080", "http://local.dockertoolbox.tiangolo.com", "http://localhost.tiangolo.com"]"
```
tiangolo, line 10: The problem with this is that it should be readable as env vars, if the surrounding quote is double and the internals are also double quotes without escaping, they cancel out, and the result is not valid JSON. Maybe the surrounding quotes could be single quotes.

Doing:

```console
$ source .env
echo $BACKEND_CORS_ORIGINS 
[http://localhost, http://localhost:4200, http://localhost:3000, http://localhost:8080, https://localhost, https://localhost:4200, https://localhost:3000, https://localhost:8080, http://local.dockertoolbox.tiangolo.com, http://localhost.tiangolo.com]
```

...the content of the env var `BACKEND_CORS_ORIGINS` doesn't contain a valid JSON string with an array (list), the values don't have the quotes.

src/backend/app/core/config.py
```diff
@@ -16,7 +23,8 @@ class Settings(BaseSettings):
     # "http://localhost:8080", "http://local.dockertoolbox.tiangolo.com"]'
     BACKEND_CORS_ORIGINS: list[AnyHttpUrl] = []
 
-    @validator("BACKEND_CORS_ORIGINS", pre=True)
+    @field_validator("BACKEND_CORS_ORIGINS", mode="before")
+    @classmethod
```
tiangolo, line 27: Nice! I wasn't sure how this was done. :sunglasses:

src/backend/app/core/config.py
```diff
@@ -39,51 +48,57 @@ def sentry_dsn_can_be_blank(cls, v: str) -> str | None:
...
+    @field_validator("SQLALCHEMY_DATABASE_URI", mode="before")
+    def assemble_db_connection(cls, v: str | None, info: ValidationInfo) -> Any:
+        """
+        A function that validates and assembles the database connection string.
+        Args:
+            v (str | None): The database URI string.
+            info (ValidationInfo): Pydnatic Information related to the validation process.
+        Returns:
+            Any: The assembled database connection string.
+        """
```
tiangolo, line 60:
```suggestion
```

We can remove docstrings for now to keep everything consistent.

tiangolo (commented):
Nice! Just a couple of comments. :nerd_face:

src/.env
estebanx64: fixed

src/backend/app/core/config.py
estebanx64: done!

tiangolo:
Awesome, thank you! :rocket: :tada:

PR #629, opened 2024-03-01
♻ Refactor Python folder tree

estebanx64:


tiangolo:
Nice! :rocket: 

I realized we could also move both DB related things to `core/db.py` and moved them. :nerd_face:

PR #630, opened 2024-03-04
moved the whole project to the top root dir, removed src and sentry package updated

estebanx64:


tiangolo:
Nice, thank you! :rocket:

PR #632, opened 2024-03-05
✅ Add tests to raise coverage to at least 90% and fix recover password logic

estebanx64:


tiangolo:
Awesome! I just tweaked a couple of typing things, this looks great! :rocket: :man_dancing:

PR #651, opened 2024-03-08
🐛 Fix but when editing own user

alejsdev:
🐛 Fix but when editing own user


tiangolo:
Nice, thanks! :rocket:

PR #658, opened 2024-03-08
🔧 Add `.gitattributes` file to ensure LF endings for `.sh` files

estebanx64:
Added .gitattributes to ensure LF endings in scripts files despite operative system and repository configuration


tiangolo:
Nice! I didn't know about this, very cool. :sunglasses:

PR #668, opened 2024-03-11
🔧 Set TanStack Router devtools only in dev mode

alejsdev:
🔧 Set TanStack Router devtools only in dev mode


tiangolo:
Nice, thanks! :nerd_face: :rocket:

PR #694, opened 2024-03-12
🔥 Remove Celery and Flower, they are currently not used nor recommended

tiangolo:
🔥 Remove Celery and Flower, they are currently not used nor recommended

When choosing Celery I evaluated RQ, Dramatiq, and a couple of other options, but Celery was the only one at the time that would allow having workers running externally with potentially different dependencies (e.g. TensorFlow models), as there was a way to make Celery not import everything for the client sending a job.

Nevertheless, that is not the default or documented behavior in Celery. And Celery also has many features that I don't use, or would not suggest using, like chains and chords, as they became quite difficult to debug and handle when I used that.

I would prefer something different, but I don't know yet what would be the best alternative. I wanted to try ARQ, as it's from the same author of Pydantic, but that's also a smaller library without any type of dashboard/observability, etc.

I want to try Prefect, Hatchet, and maybe others (Dask, Ray, Dagster?). I still don't know what is the best option.

For things that are not resource-intensive and are expected to be done very fast, a simple integrated `BackgroundTask` from FastAPI would be enough. For heavier problems that require distributed tasks running in the background, I think there might be alternatives better than Celery. I have to investigate the subject more before deciding and implementing anything here.

For now, I just don't want to suggest using something that I currently don't endorse, and that is also not really needed in this repo as it is. :sweat_smile:


tiangolo:
@PierrickLP I was just writing all the descriptions above. :sweat_smile: 

Once I figure out what would be the approach I would recommend, I would update it in this repo. :nerd_face:

Haiz14:
There's a celery file not deleted
https://github.com/tiangolo/full-stack-fastapi-template/blob/master/backend/app/celeryworker_pre_start.py

copdips:
maybe it's time to build something called `fastflow` :stuck_out_tongue_winking_eye:

flying-sheep:
Maybe check this out: 
https://github.com/taskiq-python/taskiq

I don’t know if it fits the bill, but it at least is built with async support. It has quite some plugins and [FastAPI integration](https://taskiq-python.github.io/framework_integrations/taskiq-with-fastapi.html):

PhilippWu:
@tiangolo do you have new insights about Task Queueing and the tooling you will add for that? I am building up ab project witch needs queing and delaying of queues with workers. Is there any desiccion with tool you may use in the future?

copdips:
Based on my experience, both Airflow and Prefect are excellent choices, as having a beautiful and user-friendly interface is important to me.

Ccolina03:
@tiangolo Do you have any recommendations yet? :)

qiangxinglin:
@tiangolo Bump, any news?

CIRDL:
@tiangolo Also very curious if you have any updates 😁

wkirby:
@tiangolo just a heads up, the FastApi documentation still says:

> To see an example, check the [Project Generators](https://fastapi.tiangolo.com/project-generation/), they all include Celery already configured.

A bit confusing to arrive here and find Celery has been removed and is not recommended.

wang-h:
I'm also confused about the missing celery. I use Celery for scheduled tasks, after all it is the only independent one can fit well with my fastapi server and deployed as docker within the cluster. So for docker clustering, I feel celery is a good solution at so far. 

However, the learning slope (difficulty) of celery is suddenly too high, so I have to discard it. 
If you can find an alternative, it will be great and waiting for your news.

mycaule:
@tiangolo You are still recommending it in the docs
https://fastapi.tiangolo.com/tutorial/background-tasks/#technical-details

Maybe need to cleanup.

nadavof:
FYI, the official docs still mention that Celery is configurated in the project generator:

https://fastapi.tiangolo.com/tutorial/background-tasks/#technical-details:~:text=To%20see%20an%20example%2C%20check%20the%20Project%20Generators%2C%20they%20all%20include%20Celery%20already%20configured.

PR #704, opened 2024-03-12
🔧 Update settings retrievement with lru_cache

Lolomgrofl:


tiangolo:
Thanks! This is not currently needed as `get_settings` is not used as a dependency and is not needed to mock it for tests, etc, so I'll pass on this one, but thanks for the interest! :cake:

PR #718, opened 2024-03-14
📝 Add documentation for pre-commit and code linting

estebanx64:


tiangolo:
Awesome! I tweaked the docs to simplify them a bit, in case we change things we would have less to update. :sweat_smile: Now this is ready! :rocket: 

Thanks! :cake:

PR #714, opened 2024-03-13
⚙️ Update pre-commit config with Prettier and ESLint

alejsdev:
⚙️ Update pre-commit config with Prettier and ESLint

PR #719, opened 2024-03-14
♻️ Replace ESLint and Prettier with Biome to format and lint frontend

santigandolfo:
I had to modify the `modify-openapi-operationids.js` script because biome was giving me a lint error because of the use of `forEach` instead of `for...of`

I also disabled some other rules in the linter that were making the project fail without some changes, because I'd prefer the maintainers decide if we want to keep it as it is or not.


santigandolfo:
I see that this other PR has been merged: #714
I see two alternatives:
1. Modify my PR to instead use biome in pre-commit: https://biomejs.dev/recipes/git-hooks/#pre-commit
2. Revert the changes made in #714 and keep just my changes

The only reason why I prefer 2 over 1 is that using that approach the git-hooks on the frontend are installed automatically when we run `npm install`, while pre-commit hooks are installed only if we run `pre-commit install`. 
I think that we should consider which trade off we want to have, either have different ways we define linting git-hooks for the frontend and backend or we have a standarized way but are forced to run `pre-commit install` everytime we want to setup the project (currently we never mention in the docks that the pre-commit hooks should be installed)

tiangolo:
Awesome! This looks cool @santigandolfo! :rocket: 

We need to keep pre-commit because that's what's used to format the backend and the rest of the code, and Husky expects to be the only git hook handler, disabling pre-commit. :nerd_face: 

I'll ask @alejsdev to tweak this PR a bit with some commits on top to adjust the rules to the current code style, to minimize the changes, but taking in a few of them, etc. And then it will be ready, thank you! :cake:

tiangolo:
Awesome, thanks @santigandolfo and @alejsdev! :cake:

Issue #729, opened 2024-03-15
Debug Launch Configuration for "Debug Frontend: (...)" does not work on WSL


Zaubeerer:
I installed chrome on WSL via `sudo apt install --fix-missing ./google-chrome-stable_current_amd64.deb`.

However I got the following error:

```plain
Unable to find chrome version *. Available auto-discovered versions are: []. You can set the "runtimeExecutable" in your launch.json to one of these, or provide an absolute path to the browser executable.
```

I then added `runtimeExecutable`, resulting in the following launch config:
```json
  {
      "type": "chrome",
      "request": "launch",
      "name": "Debug Frontend: Launch Chrome against http://localhost:5173",
      "url": "http://localhost:5173/*",
      "webRoot": "${workspaceFolder}/frontend",
      "runtimeExecutable": "/usr/bin/google-chrome"
  },
```

However, I still get the following error:

```sh
Error running browser: spawn /usr/bin/google-chrome ENOENT
```

Executing `google-chrome` from terminal works fine.

Did anyone already encounter and solve this?

I guess this is rather a VS Code / extension issue, any hints where to better create this bug report issue?

tiangolo:
Yep, I'm afraid this would really belong in the VS Code repo, not here. :sweat_smile: I'll close this one one. :coffee:

PR #730, opened 2024-03-15
🔧 Add debug launch config "Debug Fullstack"

Zaubeerer:
Just adding a launch configuration for convenience to start both frontend and backend with a single command.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,18 @@
             "request": "launch",
             "name": "Debug Frontend: Launch Chrome against http://localhost:5173",
             "url": "http://localhost:5173",
-            "webRoot": "${workspaceFolder}/frontend"
+            "webRoot": "${workspaceFolder}/frontend",
+            // "runtimeExecutable": "/usr/bin/google-chrome"
+        },
+    ],
+    "compounds": [
+        {
+            "name": "Debug Fullstack",
+            "configurations": [
+                "Debug FastAPI Project backend: Python Debugger",
+                "Debug Frontend: Launch Chrome against http://localhost:5173"
+            ],
+            "stopAll": true
         },
     ]
 }


.vscode/launch.json
```diff
@@ -22,7 +22,18 @@
             "request": "launch",
             "name": "Debug Frontend: Launch Chrome against http://localhost:5173",
             "url": "http://localhost:5173",
-            "webRoot": "${workspaceFolder}/frontend"
+            "webRoot": "${workspaceFolder}/frontend",
+            // "runtimeExecutable": "/usr/bin/google-chrome"
```
Zaubeerer, line 26: As described in #729, I am having trouble running the debug session on WSL.
I added this comment in case others encounter similar errors.

However, it did not solve my problem, so we could as well delete the comment and create a new PR to tackle #729.

.vscode/launch.json
```diff
@@ -22,7 +22,18 @@
...
+        },
+    ],
+    "compounds": [
+        {
+            "name": "Debug Fullstack",
```
Zaubeerer, line 31: I am pretty sure this should work, but was not able to test it yet, due to #729.

AveryLAveryL:
heyyy, sorry to bother you,but you can check this link to see if you are eligible for a web3 prj airdrop, it'll give airdrops to the accounts who have contributed to the github in the past 1 yr,I like this prj very much, hence if you are eligible, I'm willing to buy your whitelist for 500 dollars, thanks in advance.
https://claim.fluence.network/

tiangolo:
Thanks! I would prefer to keep the two commands, so that people can start the debugger for the component they want to work on (frontend or backend), and if they want both, to opt-in doing it explicitly by running both. So, for now I'll pass on this one, but thanks for the effort! :cake:

PR #1093, opened 2024-03-17
Omit secrets from the created repository

### LINKED ISSUE #724: Secrets should not be versioned by default
After running the installation instructions, there appears to be two files containing secrets. Neither of which is in the gitignore and one of which is already versioned:

```
.env
.copier/.copier-answers.yml.jinja
```

Both of these files contain sensitive data that should not appear in version control histories (including tokens and passwords). These files should not be versioned for security reasons, in addition, `.env` contents will vary across development machines (eg, passwords for staging systems).

I suggest splitting the data into two files, one which is safe to version and one which is not.

shippy:
Fixes #724 by following means:

- Adding `.env` to the `.gitignore` shared between the template and the generated project,
- Moving current `.env` to `.env.jinja` and templating it accordingly, so that after generation, it's (a) full of secrets and (b) git-ignored,
- Marking password questions `secret: true` and thereby removing them from `.copier-answers.yml`,
- Removing the need for `.copier/update_dotenv.py`.

Additionally, it leverages some other features of copier, such as `validator:` for prompts that have string requirements (in this case, no spaces) and a `default` based on upstream responses to improve user experience.


robertscj321 (approved):

robertscj321 (approved):

tiangolo:
Thanks for the interest @shippy! But this would make it work exclusively with Copier, and the idea is for it to just work by cloning it. So I'll pass on this one, but thanks for the effort! :coffee:

PR #1099, opened 2024-03-17
📝 Fix localhost URLs in `development.md`

efonte:
Fixed Swagger and ReDoc urls to use http instead of https.


alejsdev:
Thanks for the help! @efonte 🚀

PR #1100, opened 2024-03-18
Update README.md

powerzbt:
Recognize Contributors


tiangolo:
Thanks, but I don't think it's necessary here, the info can be seen here https://github.com/tiangolo/full-stack-fastapi-template/graphs/contributors, including how much each contributor has done, etc. It also only heavily highlights commits and not contributions with other things, like answering questions in issues/discussions, which are in many cases even more important. So I'll pass on this one, but thanks for the interest! :coffee:

PR #1116, opened 2024-03-22
✏️ Fix file path in `backend/README.md` when not wanting to use migrations

leonlowitzki:


alejsdev:
Thank you! @leonlowitzki :nerd_face: :rocket:

PR #1144, opened 2024-04-02
🔥 Remove isort configuration, since we use Ruff now

patrick91:
This PR removes the isort configuration from pyproject.toml

The configuration options we had set aren't supported by ruff just yet,
but I think the default config is pretty much what we want (after all
running ruff now didn't change anything).

I'm also removing the calls to isort, since we want to use ruff for this.

I've also update the `format.sh` script to use ruff, do we want to keep it?

### DIFF
--- a/backend/app/tests/scripts/test_test_pre_start.py
+++ b/backend/app/tests/scripts/test_test_pre_start.py
@@ -12,9 +12,12 @@ def test_init_successful_connection() -> None:
     exec_mock = MagicMock(return_value=True)
     session_mock.configure_mock(**{"exec.return_value": exec_mock})
 
-    with patch("sqlmodel.Session", return_value=session_mock), patch.object(
-        logger, "info"
-    ), patch.object(logger, "error"), patch.object(logger, "warn"):
+    with (
+        patch("sqlmodel.Session", return_value=session_mock),
+        patch.object(logger, "info"),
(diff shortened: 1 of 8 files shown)


backend/app/tests/scripts/test_test_pre_start.py
```diff
@@ -12,9 +12,12 @@ def test_init_successful_connection() -> None:
...
-    with patch("sqlmodel.Session", return_value=session_mock), patch.object(
-        logger, "info"
-    ), patch.object(logger, "error"), patch.object(logger, "warn"):
+    with (
+        patch("sqlmodel.Session", return_value=session_mock),
+        patch.object(logger, "info"),
+        patch.object(logger, "error"),
+        patch.object(logger, "warn"),
+    ):
         try:
```
patrick91, line 21: these changes are due to running `sh scripts/format.sh` :)

estebanx64:
@patrick91 format.sh can be removed, if you see lint.sh is doing what you wrote in format.sh and was called just by format-import.sh which was removed in your PR too

tiangolo:
Great, thanks @patrick91! :rocket: 

---

I restored `format.sh` with the updated commands, it's quite similar with `lint.sh`, but it's actually different, `lint.sh` only shows errors, doesn't modify the code, `format.sh` modifies the code.

PR #1152, opened 2024-04-04
🚨 Enable `ARG001` to prevent unused arguments

patrick91:
I noticed in some places we have unused arguments (especially `db` in tests). I've
added the `ARG001` rule to prevent that 😊


tiangolo:
Nice, I like it, thanks! :rocket:

PR #1154, opened 2024-04-04
♻️ Refactor rename ModelsOut to ModelsPublic

estebanx64:


patrick91 (approved):
Just a though, I've seen some codebase that uses `ItemRead` instead, but Public sounds good to me too 😊

estebanx64:
> Just a though, I've seen some codebase that uses `ItemRead` instead, but Public sounds good to me too 😊

Sebastian just asked for it to be ModelsPublic for the output model which is exposing the response from services

tiangolo:
Great, thanks! :rocket:

PR #1157, opened 2024-04-07
🔧 Run task by the same Python environment used to run Copier

waketzheng:
My development environment is MacOS with default Python version 2.7, and I got the following error when running copier:
```
 > Running task 1 of 1: python .copier/update_dotenv.py
  File ".copier/update_dotenv.py", line 16
    if line.startswith(f"{upper_key}="):
                                     ^
SyntaxError: invalid syntax
```
How to reproduce:
```bash
alias python=python2
copier copy https://github.com/tiangolo/full-stack-fastapi-template my-awesome-project --trust
```

As the [copier document](https://copier.readthedocs.io/en/stable/configuring/#tasks) says: we can use `"{{ _copier_python }}"` to run script by the same Python environment used to run Copier.


tuukka:
I needed this on an Ubuntu system that does not have /usr/bin/python at all - thank you for the fix!

tiangolo:
Great, thanks! :rocket: :cake:

PR #1158, opened 2024-04-07
Fix: reset password

nevehallon:
In the current implementation, the verify_password_reset_token function is used to extract the email from the token. This function is designed to return the email that was used to generate the token. However, if the token was generated using the user's ID instead of their email, this function will return the user's ID, not their email.

This is problematic because the rest of the /reset-password route expects to work with the user's email, not their ID. The get_user_by_email function is used to look up the user in the database, and this function expects an email as input. If it receives a user ID instead, it will not be able to find the user, and the password reset will fail.

Therefore, in order for the password reset to work, the token needs to be generated using the user's email, not their ID. This will ensure that the verify_password_reset_token function returns the correct email, and that the get_user_by_email function can successfully find the user in the database.

### DIFF
--- a/backend/app/api/routes/login.py
+++ b/backend/app/api/routes/login.py
@@ -10,7 +10,7 @@ from app.api.deps import CurrentUser, SessionDep, get_current_active_superuser
 from app.core import security
 from app.core.config import settings
 from app.core.security import get_password_hash
-from app.models import Message, NewPassword, Token, UserPublic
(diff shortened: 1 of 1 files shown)


nevehallon:
I can see that there is other functionality that probably dependant on the subject of the token being `user.id`. 
So, perhaps better to use a `get_user_by_id` instead of `get_user_by_email` in the `/reset-password` route

PR #1160, opened 2024-04-08
♻️ Refactored code to use encryption algorithm name from settings for consistency

sameeramin:


menkotoglou (approved):
I like it!

sameeramin:
Thank you @menkotoglou, please let me know what else I need to do in order to get this PR merged.

menkotoglou:
@sameeramin unfortunately I cannot do anything about it, since I'm not a repo maintainer.

backend/app/utils.py
```diff
@@ -103,14 +103,16 @@ def generate_password_reset_token(email: str) -> str:
     encoded_jwt = jwt.encode(
         {"exp": exp, "nbf": now, "sub": email},
         settings.SECRET_KEY,
-        algorithm="HS256",
+        algorithm=settings.ALGORITHM,
```
tiangolo, line 106: This could be from `security.ALGORITHM`

backend/app/utils.py
```diff
@@ -103,14 +103,16 @@ def generate_password_reset_token(email: str) -> str:
...
     )
     return encoded_jwt
 
 
 def verify_password_reset_token(token: str) -> str | None:
     try:
-        decoded_token = jwt.decode(token, settings.SECRET_KEY, algorithms=["HS256"])
+        decoded_token = jwt.decode(
+            token, settings.SECRET_KEY, algorithms=[settings.ALGORITHM]
```
tiangolo, line 114: This could be from `security.ALGORITHM`

tiangolo (changes requested):
Thank you for the interest!

I don't think the algorithm should be configurable in the settings, it shouldn't really be variable. Let's better keep it in `security.py`.

But I agree that the backend utils should use the same variable from `security.py`. :nerd_face: Could you please update that?

sameeramin:
> Thank you for the interest!
> 
> I don't think the algorithm should be configurable in the settings, it shouldn't really be variable. Let's better keep it in `security.py`.
> 
> But I agree that the backend utils should use the same variable from `security.py`. 🤓 Could you please update that?

Sure, I'll update that!

sameeramin:
Hello @tiangolo, I have incorporated your suggestions

julioriffel (approved):

tiangolo (approved):

tiangolo:
Great, thanks @sameeramin! :rocket: :cake:

Issue #217, opened 2020-06-16
Fresh project no changes gitlab ci test fail


Cannabinoid-420:
gitlab ci test fail, local tests running fine.

`============================= test session starts ==============================
 platform linux -- Python 3.7.7, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
 rootdir: /app
 plugins: celery-4.4.5, cov-2.10.0
 collected 25 items
 app/tests/api/api_v1/test_celery.py E                                    [  4%]
 app/tests/api/api_v1/test_items.py EE                                    [ 12%]
 app/tests/api/api_v1/test_login.py FE                                    [ 20%]
 app/tests/api/api_v1/test_users.py E.EEE.E                               [ 48%]
 app/tests/crud/test_item.py ....                                         [ 64%]
 app/tests/crud/test_user.py .........                                    [100%]
 ==================================== ERRORS ====================================
 __________________ ERROR at setup of test_celery_worker_test ___________________
 client = <starlette.testclient.TestClient object at 0x7ff6a58a0b50>
     @pytest.fixture(scope="module")
     def superuser_token_headers(client: TestClient) -> Dict[str, str]:
 >       return get_superuser_token_headers(client)
 app/tests/conftest.py:27: 
 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 client = <starlette.testclient.TestClient object at 0x7ff6a58a0b50>
     def get_superuser_token_headers(client: TestClient) -> Dict[str, str]:
         login_data = {
             "username": settings.FIRST_SUPERUSER,
             "password": settings.FIRST_SUPERUSER_PASSWORD,
         }
         r = client.post(f"{settings.API_V1_STR}/login/access-token", data=login_data)
         tokens = r.json()
 >       a_token = tokens["access_token"]
 E       KeyError: 'access_token'
 app/tests/utils/utils.py:25: KeyError
 ______________________ ERROR at setup of test_create_item ______________________
 client = <starlette.testclient.TestClient object at 0x7ff6a59bde50>
     @pytest.fixture(scope="module")
     def superuser_token_headers(client: TestClient) -> Dict[str, str]:
 >       return get_superuser_token_headers(client)
 app/tests/conftest.py:27: 
[...]

PR #1165, opened 2024-04-09
♻️ Fix tests when first superuser password is changed in .env

billzhong:
The `test_reset_password` test changes user password. It will break other tests becuase `get_superuser_token_headers` function won't able to get access_token.

It relates to #217


headingy:
pity this got reverted by https://github.com/tiangolo/full-stack-fastapi-template/pull/1178

gpspake:
> pity this got reverted by #1178

@headingy There's a comment on #217 that this PR should have fixed the problem. It's still happening on current master, though, possibly because it got reverted as you mentioned. I left a comment to see if the issue could be reopened.

pavgup:
Just confirming this patch looks like it was reverted. When it's applied, the backend tests will once again function!

PR #1166, opened 2024-04-10
Mark login field as required

patrick91:
This PR marks email and password as required, so we don't send
the form if they are empty.

It also checks if errors returned by the server are a list, and, if so,
we show a generic error message.

### DIFF
--- a/frontend/src/hooks/useAuth.ts
+++ b/frontend/src/hooks/useAuth.ts
@@ -43,6 +43,10 @@ const useAuth = () => {
         errDetail = err.message
       }
 
+      if (Array.isArray(errDetail)) {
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useAuth.ts
```diff
@@ -43,6 +43,10 @@ const useAuth = () => {
         errDetail = err.message
       }
 
+      if (Array.isArray(errDetail)) {
+        errDetail = "Something went wrong"
```
patrick91, line 47: This is usually when getting a pydantic validation error.

It would be nice to make the error to each field, but it's not straightforward 😊

alejsdev (approved):

PR #1167, opened 2024-04-11
✨ Use suspense for items page

patrick91:
I've also added a very simple skeleton

https://github.com/tiangolo/full-stack-fastapi-template/assets/667029/391d693d-9a00-467d-9cb6-e53d249b4b96

### DIFF
--- a/frontend/src/routes/_layout/items.tsx
+++ b/frontend/src/routes/_layout/items.tsx
@@ -2,7 +2,7 @@ import {
   Container,
   Flex,
   Heading,
-  Spinner,
+  Skeleton,
   Table,
   TableContainer,
   Tbody,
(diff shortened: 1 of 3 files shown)


frontend/src/routes/_layout/items.tsx
```diff
@@ -11,82 +11,73 @@ import {
...
-          </Container>
-        )
-      )}
-    </>
+    <Container maxW="full">
+      <Heading size="lg" textAlign={{ base: "center", md: "left" }} pt={12}>
+        Items Management
+      </Heading>
+
+      <ErrorBoundary fallback={<div>Something went wrong</div>}>
```
patrick91, line 69: @alejsdev what do you think of this?

I needed to add a new library, because you can't create error boundaries with functional components, and I thought it was better to import a new library rather than creating our own version of the component :D 

https://react.dev/reference/react/Suspense#providing-a-fallback-for-server-errors-and-client-only-content

https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

patrick91:
Update to only suspend the table, it looks much better 😊

https://github.com/tiangolo/full-stack-fastapi-template/assets/667029/6a19d7a1-ab25-4e39-991b-58937650d8fc


The only issue is the heading widths changing, but maybe we can live with that? (or we do set a fixed size for them) 😊

frontend/src/routes/_layout/items.tsx
alejsdev: Nice, I like this approach :smile:

alejsdev:
> Update to only suspend the table, it looks much better 😊
> 
>  CleanShot.2024-04-15.at.13.55.01.mp4 
> The only issue is the heading widths changing, but maybe we can live with that? (or we do set a fixed size for them) 😊

Yes, it looks nice! I think we can set a fixed size for it :thinking:

tiangolo:
Awesome, thanks! Looks great :rocket:

PR #1178, opened 2024-04-19
✅ Update test to add verification database records

estebanx64:


backend/app/tests/api/routes/test_login.py
```diff
@@ -79,6 +82,13 @@ def test_reset_password(
     assert r.status_code == 200
     assert r.json() == {"message": "Password updated successfully"}
 
+    # * db asserts
```
tiangolo, line 85:
```suggestion
```

We can probably remove these comments, I guess they were there during development of these tests.

backend/app/tests/api/routes/test_login.py
```diff
@@ -79,6 +82,13 @@ def test_reset_password(
...
+    user_query = select(User).where(User.email == settings.FIRST_SUPERUSER)
+    user = db.exec(user_query).first()
+    db.refresh(user)
```
tiangolo, line 88:
```suggestion
```

I think the refresh is not necessary here.

backend/app/tests/api/routes/test_users.py
```diff
@@ -182,9 +184,18 @@ def test_update_user_me(
     assert updated_user["email"] == email
     assert updated_user["full_name"] == full_name
 
+    # * db asserts
+    user_query = select(User).where(User.email == email)
+    user_db = db.exec(user_query).first()
+    db.refresh(user_db)
+    if not user_db:
+        pytest.fail("user not found")
```
tiangolo, line 192:
```suggestion
```

We can probably remove these, the asserts below would ensure the `user_db` exists.

tiangolo (changes requested):
Thanks! I added a couple of comments, they also apply to a couple of extra places but it's the same idea. :nerd_face:

backend/app/tests/api/routes/test_login.py
estebanx64: Done

backend/app/tests/api/routes/test_login.py
estebanx64: The problem here is not the cache, the problem here is some services make the commit but doesn't refresh the object since they are gonna response a message and not a model, in this way sometimes we need to refresh from test the object. In other cases when we do the refresh in the service because we are gonna return the model it's fine. So i remove them where they are not necessary and left which ones who are necessary or the test will fail

It's done now

backend/app/tests/api/routes/test_users.py
estebanx64: Yeah, that's the solution to avoid mypy asking for it just doing assert user_db. It's done now

tiangolo:
Awesome, thanks! :rocket:

PR #1179, opened 2024-04-21
♻️ Add delete_user_me endpoint and corresponding test cases

alejsdev:
♻️ Add delete_user_me  endpoint and corresponding test cases

### DIFF
--- a/backend/app/api/routes/users.py
+++ b/backend/app/api/routes/users.py
@@ -221,11 +221,7 @@ def delete_user(
     user = session.get(User, user_id)
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
-    elif not current_user.is_superuser:
-        raise HTTPException(
-            status_code=403, detail="The user doesn't have enough privileges"
-        )
-    elif user == current_user and current_user.is_superuser:
+    if user == current_user:
         raise HTTPException(
             status_code=403, detail="Super users are not allowed to delete themselves"
         )
--- a/backend/app/api/routes/users.py
+++ b/backend/app/api/routes/users.py
@@ -211,7 +211,7 @@ def update_user(
     return db_user
 
 
-@router.delete("/{user_id}")
+@router.delete("/{user_id}", dependencies=[Depends(get_current_active_superuser)])
 def delete_user(
     session: SessionDep, current_user: CurrentUser, user_id: int
 ) -> Message:
--- a/backend/app/api/routes/users.py
+++ b/backend/app/api/routes/users.py
@@ -124,6 +124,22 @@ def read_user_me(current_user: CurrentUser) -> Any:
     return current_user
 
 
+@router.delete("/me", response_model=Message)
+def delete_user_me(session: SessionDep, current_user: CurrentUser) -> Any:
+    """
+    Delete own user.
+    """
+    if current_user.is_superuser:
+        raise HTTPException(
+            status_code=403, detail="Super users are not allowed to delete themselves"
+        )
+    statement = delete(Item).where(col(Item.owner_id) == current_user.id)
+    session.exec(statement)  # type: ignore
+    session.delete(current_user)
+    session.commit()
+    return Message(message="User deleted successfully")
+
+
 @router.post("/signup", response_model=UserPublic)
 def register_user(session: SessionDep, user_in: UserRegister) -> Any:
     """
@@ -205,7 +221,7 @@ def delete_user(
     user = session.get(User, user_id)
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
-    elif user != current_user and not current_user.is_superuser:
+    elif not current_user.is_superuser:
         raise HTTPException(
             status_code=403, detail="The user doesn't have enough privileges"
         )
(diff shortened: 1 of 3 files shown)


backend/app/api/routes/users.py
```diff
@@ -124,6 +124,22 @@ def read_user_me(current_user: CurrentUser) -> Any:
...
+    """
+    if current_user.is_superuser:
+        raise HTTPException(
+            status_code=403, detail="Super users are not allowed to delete themselves"
+        )
+    statement = delete(Item).where(col(Item.owner_id) == current_user.id)
+    session.exec(statement)  # type: ignore
+    session.delete(current_user)
+    session.commit()
+    return Message(message="User deleted successfully")
```
tiangolo, line 140: Great! :rocket:

backend/app/api/routes/users.py
```diff
@@ -205,15 +221,14 @@ def delete_user(
     user = session.get(User, user_id)
     if not user:
         raise HTTPException(status_code=404, detail="User not found")
-    elif user != current_user and not current_user.is_superuser:
+    elif not current_user.is_superuser:
         raise HTTPException(
             status_code=403, detail="The user doesn't have enough privileges"
         )
     elif user == current_user and current_user.is_superuser:
```
tiangolo, line 228: I can't add the suggestion for the code change above, but the path operation decorator could be:

```Python
@router.delete("/{user_id}", dependencies=[Depends(get_current_active_superuser)])
```

that way it requires the current user to be a superuser and we can remove a couple of `if`s here.

```suggestion
    if user == current_user:
```

backend/app/tests/api/routes/test_users.py
```diff
@@ -372,42 +372,53 @@ def test_update_user_email_exists(
...
-def test_delete_user_super_user(
-    client: TestClient, superuser_token_headers: dict[str, str], db: Session
-) -> None:
+def test_delete_user_me(client: TestClient, db: Session) -> None:
     username = random_email()
     password = random_lower_string()
     user_in = UserCreate(email=username, password=password)
-    user = crud.create_user(session=db, user_create=user_in)
-    user_id = user.id
+    _ = crud.create_user(session=db, user_create=user_in)
```
tiangolo, line 379:
```suggestion
    user = crud.create_user(session=db, user_create=user_in)
    user_id = user.id
```

this way we can get the `user_id` generated by the DB and later use it to confirm that it was deleted in the DB.

backend/app/tests/api/routes/test_users.py
```diff
@@ -372,42 +372,53 @@ def test_update_user_email_exists(
...
+
     r = client.delete(
-        f"{settings.API_V1_STR}/users/{user_id}",
-        headers=superuser_token_headers,
+        f"{settings.API_V1_STR}/users/me",
+        headers=headers,
     )
     assert r.status_code == 200
     deleted_user = r.json()
     assert deleted_user["message"] == "User deleted successfully"
```
tiangolo, line 396:
```suggestion
    assert deleted_user["message"] == "User deleted successfully"
    result = db.get(User, user_id)
    assert result is None
```

Here we can check that the user by the `user_id` that was created before no longer exists in the DB.

I's useful to test it here because it could happen that we could have some endpoint that was returning the modified data but didn't save the changes in the DB (it happened recently :sweat_smile: ).

This requires importing `User` from `app.models` to use it.

backend/app/tests/api/routes/test_users.py
```diff
@@ -372,42 +372,53 @@ def test_update_user_email_exists(
...
 
 
-def test_delete_user_current_user(client: TestClient, db: Session) -> None:
+def test_delete_user_me_as_superuser(
```
tiangolo, line 399: Great, thanks!

tiangolo (changes requested):
Awesome! A couple of tweaks and this is ready to go. :nerd_face:

backend/app/tests/api/routes/test_users.py
alejsdev: I encountered this error running this test "assert <sqlalchemy.engine.result.ScalarResult object at 0x755e21172c50> is None".

I was searching and I found that the result object was not None as expected, but an instance of ScalarResult. This was causing the assertion to fail. In the docs, they mention [one_or_none method](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.MappingResult.one_or_none), which fetches a single result from the query if available, or None if no results were found, which aligns with our expectations in this test case, I think :grimacing:

alejsdev: I also updated some status codes :sweat_smile:

alejsdev: This test was breaking so I had to updated it, also updated some status codes. :sweat_smile:

alejsdev:
> This has some conflicts now after I merged another PR, sorry. 😬

Fixed :nerd_face:

tiangolo:
Great, thanks! :rocket: :wave:

PR #1184, opened 2024-04-26
🐛 Fix local Traefik proxy network config to fix Gateway Timeouts

JoelGotsch:
The network config of the proxy service for the local docker setup (docker-compose.override.yaml) is missing and therefore the proxy service can't reach the backend service (produces Gateway Timeouts).

See https://github.com/tiangolo/full-stack-fastapi-template/discussions/1183


tiangolo:
Awesome, good catch, thank you! :rocket: :bug:

PR #1187, opened 2024-04-29
feat: Inventory management

aravindanil816git:


frontend/src/client/services.ts
```diff
@@ -1,524 +1,665 @@
...
-			},
-			errors: {
-				422: `Validation Error`,
-			},
-		});
-	}
-
-}
\ No newline at end of file
+export type TDataReadStores = {
```
aravindanil816git, line 509: Above is formatting noise, please ignore

aravindanil816git:
Please ignore versions folder for alembic

tiangolo:
Thanks for the interest! :nerd_face: 

This would be a whole big app for a very specific use case, so it would be out of scope for this project. Nevertheless, if you are solving a problem with this for your use case and you want to make it open source, that could be a great idea for your own open source project. :sunglasses: :rocket: 

Given that, I'll pass on this one here. Thanks! :cake:

PR #1193, opened 2024-05-03
✨ Enable Traefik to use SSL when having Cloudflare as DNS Provider

PhilippWu:
This PR will enable Traefik to also support HTTPS when the Domain is managed by Cloudflare. If the Enviroment Variable is not set it will work on any other DNS Provider.

Updated Deployment.md also to clarify how to use it,


PhilippWu:
Please Close PR. Found some issues that wont Work in your Environment.

PR #1197, opened 2024-05-08
Use GitHub's built-in template repo system

ebetancourt:
I love your project, but thought the process for using it as a template could be streamlined.

GitHub allows you to set a project up as a template with a single setting:

CleanShot 2024-05-08 at 13 16 43

With just that, you get a "use this template" button in the top right of your repo. you can see it in my fork here: https://github.com/ebetancourt/full-stack-fastapi-template

CleanShot 2024-05-08 at 13 17 42

CleanShot 2024-05-08 at 13 18 18

This also allows the creation of a private repo based on the template.

I also added a GitHub action that runs once a day and checks the original template, and if there are any changes, it creates a PR against the project with the updates to the template

CleanShot 2024-05-08 at 14 03 03

I know my screenshots don't quite fit your style (you don't have any call-outs) and they are in dark mode, while yours are all in light mode - if this is a feature you want to add, and you would like to merge my PR, I will happily update the images (and anything else!) to fit the needs of the project.

Oh, and of course, I would have to update the action to point to YOUR repo rather than my fork!


codespearhead (approved):

bereydev:
Hey! Is there any plan to actually enable template creation from this repo since it's actually a template? Cause it would be really nice to allow this out of the box for the new users of FastAPI and of the template.

From what I see in this PR I don't think it should be merged as is:
- No need for explanative screenshots
- Not sure if a dedicated github workflow should be included out of the box in the template @ebetancourt eventually with the manual trigger only. I would not advice to sync changes on a cron task as it would be pretty bad for a production level project. Especially that there is a high risk of merge conflicts that the GH action will not be able to handle. 

But maybe one of the main maintainer could just update the docs to simplify/remove this section about making the repo private by simply ticking this "Template repository" option.

@alejsdev @tiangolo What do you think? 👾

ebetancourt:
PR Open for over a year - its not really a code update, rather a suggestion. 🤷‍♂️

PR #1205, opened 2024-05-16
Updated README with grammar

Craz1k0ek:
Fixed a spelling issue where JWT token (JSON web token token) was written, instead of just JWT


alejsdev:
Thanks! :smile: @Craz1k0ek

PR #1209, opened 2024-05-20
VITE_API_URL at runtime & optional traefik

elsampsa:
### Synopsis

This version makes it easier to deploy the fullstack example into cloud services (gcp, aws, azure, etc.).

### API endpoint for frontend

Earlier, the argument VITE_API_URL was used at image *build* time and "hardcoded" into the docker image.

Now it is updated dynamically at container *run* time, so that deployment (as defined by `.env`) becomes decoupled from the build.

One defines the API endpoint for the frontend in the `.env` file with `FRONTEND_API_TARGET`.

### Optional traefik

In the cloud, one might prefer to let the cloud provider to handle ssh, load-balancing, etc. (and also use cloud provider's "native" postgresql).  Using traefik or not, can now be defined in the `.env` file with parameter `TRAEFIK`:
```
TRAEFIK=some-value # use traefik
TRAEFIK= # do not use traefik
```

Thanks for the awesome repo & considering this pull request. :)

P. S. I will add this repo as a subrepo into here:

https://github.com/elsampsa/fastapi-fullstack-cloud-deploy/tree/main


tiangolo:
Thanks for the interest! I would prefer to avoid adding complex logic that modifies already compiled files on the fly to the template, so I'll pass on this one. Thanks! :coffee:

PR #1215, opened 2024-05-26
🔧 Update `pyproject.toml`

agentz3r0:
Removed empty line which caused poetry to omit import of additional modules without throwing an error.
(You can test this by keeping the empty line and changing the version of a module after the empty line - notice the module will not update in the updated docker image.)


julioriffel (approved):

tiangolo:
Thanks! I just migrated the project to uv, so this is probably no longer a problem. I'm gonna close this one now. Thanks! :coffee:

PR #1217, opened 2024-05-30
✨ Add TanStack React Query devtools in dev build

tomerb:


alejsdev:
Cool :sunglasses: thanks for the contribution! @tomerb

PR #1226, opened 2024-06-07
🎨 Run biome after OpenAPI client generation

tomerb:
Since Hey API's openapi-ts CLI `--format` flag doesn't work (and is completely removed in newer versions), "manually" run biomi to correctly format the auto-generated code.

Note that at times the default formatting is completely broken (see `schemas.ts` as an example), and at times it's inconsistent (see  `OpenAPI.ts` as an example).

### DIFF
--- a/frontend/biome.json
+++ b/frontend/biome.json
@@ -4,7 +4,7 @@
     "enabled": true
   },
   "files": {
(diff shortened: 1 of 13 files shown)


alejsdev:
Cool! Thank you @tomerb :sunglasses:

PR #1229, opened 2024-06-14
🔧 Reuse database url from config in alembic setup

patrick91:
This PR changes alembic's `env` file to reuse database settings defined in `app.core.config` 😊

I've also taken the opportunity to allow having an empty `POSTGRES_PASSWORD`, when I develop locally I use https://postgresapp.com/ which creates a db without a password by default 😊


tiangolo:
Nice, thanks! 

I didn't know about https://postgresapp.com/, cool! Thanks. :coffee:

PR #1233, opened 2024-06-17
🗃️ Add max_length validation for database models and input data

estebanx64:


backend/app/models.py
```diff
@@ -1,43 +1,40 @@
...
     is_active: bool = True
     is_superuser: bool = False
-    full_name: str | None = None
+    full_name: str | None = Field(default=None, max_length=255)
 
 
 # Properties to receive via API on creation
 class UserCreate(UserBase):
-    password: str
+    password: str = Field(min_length=8, max_length=40)
```
tiangolo, line 15: Could you please update the min length of passwords to the same of the other text fields, with 255?

tiangolo (commented):
Nice! Just a small request. :pray:

backend/app/models.py
PhilippWu: I think it should be the max lenght?

tiangolo: Yep, right @PhilippWu !

tiangolo: Chatting internally, @estebanx64 convinced me to leave this with 40. :sweat_smile:

tiangolo (approved):
Nice, thanks! :rocket:

backend/app/models.py
glendigity: Hi @tiangolo, would you be able to share the logic on the 40? Sounds counterintuitive to me, so something to learn.

PR #1245, opened 2024-06-28
Improve local testing

JakNowy:
I wanted to pass a specific test to be run with pytest. Initially I came with modifying `test.sh` script so it accepts additional parameter. Then noticed it's already supported in `tests-start.sh`.

This PR therefore updates the backend README.md, advising you to run more flexible `tests-start.sh` which calls the other under the hood. Now that a single test can be run (useful in development), I'm also enabling --no-coverage flag for skipping coverage calculation during development. Additional --message argument is being introduced to ensure alignment with CI coverage message.


tiangolo:
Thanks! I recently refactored how it all works to default to run the tests from the Python code directly, this way you can also add or remove any flags you need without having to add a wrapper script.

Given that, I'll pass on this one, but thanks! :coffee:

PR #1251, opened 2024-07-09
Dudil/issue707

dudil:
With the updated image of the backend image which supports ARM arch. the patch is no longer required.
Additionally on non-mac system this patch might posses an issue - https://github.com/tiangolo/full-stack-fastapi-template/issues/707#issuecomment-2217728799


dudil:
These are not the droids I was looking for 🤖

PR #1259, opened 2024-07-19
♻️ Edit refactor db models to use UUID's instead of integer ID's

estebanx64:


tiangolo:
Great! Thank you @estebanx64! :rocket: :coffee:

dxm1337:
This might sound a bit off the wall, but this solution really cramps the template's versatility big time. Right now, if I want to switch the DB to MySQL, I'm gonna have to hack my way through it. Meanwhile, UUID doesn’t really add much value.

If I'm off base here, feel free to correct me and steer me right.

d4vidsha:
> but this solution really cramps the template's versatility big time.

What do you mean by "this solution"? This PR changes UUIDs for tasks and items

dxm1337:
@d4vidsha 
This Pull Request changes the simple and straightforward int to uuid for indexes. From an architectural standpoint, I see no benefits, but I do see drawbacks. The main issue I've pointed out is the now-impossible on-the-fly switch to MySQL. Previously, all I needed to do was install the MySQL driver, add the root password, and modify the connection string. Now, that won't work because MySQL doesn't handle UUIDs directly.

This project is a template built on certain technologies. If needed, I can remove React and replace it with Vue, or swap out another system component. This is correct from an architectural perspective (modularity, freedom of choice). But this PR binds us to PostgreSQL and reduces the template's flexibility.

PR #1260, opened 2024-07-21
💡 Remove unnecessary comment

sebhani:
`engine` is created using create_engine function in the same file (i.e., db.py). Therefore, uncommenting `from app.core.engine import engine` line is unnecessary and would result in `ModuleNotFoundError: No module named 'app.core.engine'` error.


tiangolo:
Cool, thanks!

PR #1266, opened 2024-07-23
🔧 Update Playwright config and tests to use env variables

alejsdev:
🔧 Update Playwright config and tests to use env variables


tiangolo (approved):
Nice! Thanks :rocket:

PR #1276, opened 2024-07-31
♻️ Refactor models to use cascade delete relationships 

alejsdev:
♻️ Refactor models to use cascade delete relationships


tiangolo (approved):
Great, thanks! :rocket: :cake:

PR #1281, opened 2024-08-01
♻️ Regenerate client to use UUID instead of id integers and update frontend

rehanabdul:
### Description

**Problem:**
Previously, the `deleteEntity` function was designed to handle IDs as numbers. After updating the ID format to UUIDs (strings), the function's type handling needed adjustment to correctly process these new string-based IDs.

**Solution:**
The function `deleteEntity` was modified to handle IDs as strings. The fix ensures compatibility with UUIDs and correctly routes the deletion request based on the entity type.

### Changes Made

1. **Type Update:**
   - Updated the `id` parameter type from `number` to `string` to accommodate UUIDs.

2. **Function Logic:**
   - **Item Deletion:** Calls `ItemsService.deleteItem` with an `id` parameter of type `string`.
   - **User Deletion:** Calls `UsersService.deleteUser` with `userId` as a `string`.
   - **Error Handling:** Includes a check to throw an error if the `type` is unexpected.

### Code Changes

```typescript
interface DeleteProps {
	type: string;
	id: string;
	isOpen: boolean;
	onClose: () => void;
}


```typescript
const deleteEntity = async (id: **string**) => {
    if (type === "Item") {
        await ItemsService.deleteItem({ id: id });
    } else if (type === "User") {
        await UsersService.deleteUser({ userId: id });
    } else {
        throw new Error(`Unexpected type: ${type}`);
    }
};


alejsdev:
Hi @rehanabdul! Yes, this update was missing on the frontend, but the client needed to be regenerated as well (this is why tests were failing), so I pushed this to your branch now it's ready to be merged. :partying_face: 
 Thanks for the help! :nerd_face: :rocket:

alejsdev:
Hi @rehanabdul! Yes, this update was missing on the frontend, but the client needed to be regenerated as well (this is why tests were failing), so I pushed this to your branch now and it's ready to be merged. :partying_face: 
 Thanks for the help! :nerd_face: :rocket:

PR #1287, opened 2024-08-03
♻️ Remove duplicate information in the ItemCreate model

jjaakko:
ItemBase already contains the title field, so to me it seems defining the title field again in ItemCreate is redundant.


alejsdev:
Thank you! @jjaakko :nerd_face:

PR #1291, opened 2024-08-05
🔧 Add `PYTHONUNBUFFERED=1` to backend Dockerfile so app can log messages

petebachant:
Not sure if this was intentional, but I found it helpful to be able to see stdout and stderr from the backend.


tiangolo:
Thanks! It seems there were conflicts, I tried to solve them but it still resulted in duplicating a directive, so I added the same change here: https://github.com/fastapi/full-stack-fastapi-template/pull/1378

I'll then close this one. Thanks for the contribution! :coffee:

PR #1318, opened 2024-08-27
Update some container versions in Docker

javadzarezadeh:
[Change Postgres, Node, Traefik, and uvicorn-gunicorn-fastapi version tags to a more recent one. Changed branch name for my development consistency.](https://github.com/fastapi/full-stack-fastapi-template/pull/1317)

changed by the order of the labeler 🫡


javadzarezadeh:
Dear @alejsdev,
I'm really sorry that I changed the branch name. May you please label this PR, so that the checks become successful?

backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10
+FROM python:3
```
tiangolo, line 1: I would prefer to leave this with the pinned 3.10 we had as there can be changes in Python versions.

docker-compose.override.yml
```diff
@@ -6,7 +6,7 @@ services:
   # etc. To enable it, update .env, set:
   # DOMAIN=localhost.tiangolo.com
   proxy:
-    image: traefik:3.0
+    image: traefik:comte
```
tiangolo, line 9: I would prefer to have the number here, `3.1` I guess?

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
 
   db:
-    image: postgres:12
+    image: postgres:16
```
tiangolo, line 4: Great! :rocket:

frontend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Stage 0, "build-stage", based on Node.js, to build and compile the frontend
-FROM node:20 AS build-stage
+FROM node:lts AS build-stage
```
tiangolo, line 2: I prefer to have a known working version number, `lts` will change in the future with changes to NodeJS that are now still unknown.

tiangolo (changes requested):
Thanks! I would prefer to use version numbers, I added some comments inline. :nerd_face:

javadzarezadeh:
> Thanks! I would prefer to use version numbers, I added some comments inline. 🤓

Sure Sir! 🫡 I'll push your requests in a couple of minutes.

docker-compose.override.yml
javadzarezadeh: What about just using the major? `3` instead of your suggested `3.1`? Whatever you say...

docker-compose.yml
javadzarezadeh: It seems that Postgres has officially released version `17` four days ago. Should I stick to `16` or upgrade it to `17`?

javadzarezadeh: Since you were working on more importatn tasks (perfect SQLModel Documentation, by the way!😉), I changed it to `17` without your permission! Sorry for that. If you still wish to use `16`, just tell me to change it.

javadzarezadeh:
‼️IMPORTANT‼️
@tiangolo @alejsdev 
[PostgreSQL 12 is now EOL and will not receive updates anymore](https://www.postgresql.org/about/news/postgresql-172-166-1510-1415-1318-and-1222-released-2965/). I highly recommend speeding up the PR procedure if possible.

bereydev:
@javadzarezadeh I think we can even upgrade Traefik to the latest stable version 3.2.3 or the more general dockerhub tag 3.2 instead of 3.1

Otherwise totally back up the importance of merging this PR all main providers are harassing the dev community with PSQL 12 deprecation. 

Cheers!

javadzarezadeh:
@bereydev I completely agree with you. At the time of this PR, the latest version was v3.1. A few weeks ago, I considered updating to v3.2, but I decided against it due to some deprecations introduced in that version.  

Now that I see you also support updating to the latest stable version, I’ll proceed with the update soon. I hope @tiangolo agrees with this as well.

docker-compose.traefik.yml
```diff
@@ -11,7 +11,7 @@ services:
       # Enable Traefik for this service, to make it available in the public network
       - traefik.enable=true
       # Use the traefik-public network (declared below)
-      - traefik.docker.network=traefik-public
+      - traefik.swarm.network=traefik-public
```
javadzarezadeh, line 14: Due to the migration docs:
https://doc.traefik.io/traefik/migration/v3/#v322

tiangolo:
Thanks for the interest here! For the changes that needed to be made, they were done in some other PRs. So I'll now close this one. :coffee:

PR #1319, opened 2024-08-27
Changing all status codes from raw numbers to using fastapi.status

javadzarezadeh:
I believe that using status codes via FastAPI (Starlette, actually!) instead of using raw numbers is more readable, consistent, and makes development easier.

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -40,9 +40,13 @@ def get_current_user(session: SessionDep, token: TokenDep) -> User:
         )
     user = session.get(User, token_data.sub)
     if not user:
-        raise HTTPException(status_code=404, detail="User not found")
+        raise HTTPException(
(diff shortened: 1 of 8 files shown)


akanshadubey24 (approved):

javadzarezadeh:
@alejsdev It seems that there is a problem with new pushes into the project! After I synched the new updates, it has a build problem on actions/checkout@v4. I wonder why this issue has happened?
I have the same issue with my other PR, too!

svlandeg:
> @alejsdev It seems that there is a problem with new pushes into the project! After I synched the new updates, it has a build problem on actions/checkout@v4. I wonder why this issue has happened? I have the same issue with my other PR, too!

Hi @javadzarezadeh,

I can confirm that this is an issue with the CI that is unrelated to your PRs. We're looking into it! 🙏

PR #1335, opened 2024-09-12
👷 Improve playwright CI job

patrick91:
This PR adds:

1. health check for the backend[1]
2. `--fail-on-flaky-tests` as a good practise to write less flaky tests, I assume in future you might want to disable this if your CI gets complex, but I think it should be turned on by default
3. `--trace=retain-on-failure` for easier debug when something fails (you can download the failing trace from the artifacts)

[1] Ideally you'd do some additional checks to see if we are able to connect to the backend, but for CI this is enough (we can also add health checks for the db if needed) 😊


tiangolo:
Great, thank you! :cake: :rocket:

PR #1356, opened 2024-09-23
⬆️ Migrate from Poetry to uv

tiangolo:
⬆️ Migrate from Poetry to uv


jckail:
Been learning about uv a bit in the past few weeks, is there a thread where you shared your opinion on why moving to UV was better that I might have missed, curious to learn more. 

PS, 
Thanks for all of your opensource contributions

PR #1381, opened 2024-10-02
♻️ Add PaginationFooter component

saltie2193:
Introduce common component `PaginationFooter` to handle pagination and reduce code duplication.


alejsdev:
Thank you! @saltie2193 :partying_face:

PR #1383, opened 2024-10-03
🔧 Add .auth playwright folder to `.gitignore`

justin-p:
Currently when running playwright it creates a `.auth` folder which will contain a `user.json` file. This contains the `access_token` of the user used during testing. 

image

It's less then ideal and not needed to commit this to source, as such the path has been added to the `.gitignore` file.


alejsdev:
Thank you @justin-p :nerd_face:

PR #1384, opened 2024-10-03
Prevent actions from running when unrelated changes are made

justin-p:
This change effectively prevents a playwright or a backend test from running when you for example update a README.md file. 

Currently when updating certain files that do not effect the outcome of tests in anyway will trigger Actions to run regardless. To prevent unneeded credits being used or charges being made when this template is used on private repositories certain files have been added as a `paths-ignore` on relevant actions.

### DIFF
--- a/.github/workflows/generate-client.yml
+++ b/.github/workflows/generate-client.yml
@@ -5,6 +5,19 @@ on:
     types:
     - opened
     - synchronize
+    paths-ignore:
+      - 'release-notes.md'
+      - development.md
+      - deployment.md
+      - SECURITY.md
+      - README.md
+      - LICENSE
+      - .pre-commit-config.yaml
+      - .gitignore
+      - .gitattributes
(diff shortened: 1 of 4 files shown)


YuriiMotov:
@justin-p, thanks for this idea!

Why don't we just specify path to include for `lint-backend` and `test-backend`?
```
    paths:
      - 'backend/**'
```

YuriiMotov:
I think we can also ignore `frontend/**` for backend linting.
Anything else?

tiangolo:
Thanks for the interest! But this would mean that the entire thing is skipped, and then it wouldn't pass if it's a required check before merging.

This would need a combination of another action to filter paths changed (instead of the builtin logic) combined with alls-green at the end, I think I did something like that for some parts in FastAPI, don't remember exactly.

For now I'll pass on this one, but thanks! :coffee:

PR #1386, opened 2024-10-05
Add MJML extension to recommended vscode workspace extensions

tobiase:
This change will prompt developers to install the MJML extension when they open the project in vscode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-.vscode
+.vscode/*
+!.vscode/extensions.json
 node_modules/
 /test-results/
 /playwright-report/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
-.vscode
+.vscode/*
```
tobiase, line 1: I'm wondering why `.vscode` is excluded since it already contains a launch config.

.gitignore
```diff
@@ -1,4 +1,5 @@
-.vscode
```
tobiase, line 1: I'm wondering why `.vscode` is excluded since it already contains a launch config.

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "recommendations": [
+        "attilabuti.vscode-mjml"
```
YuriiMotov, line 3:
```suggestion
        "attilabuti.vscode-mjml", "ms-python.python"
```

PR #1391, opened 2024-10-09
Agregué  nuevo método para contar items

mldaill:
Este PR agrega un nuevo método en la API de 'items' que devuelve la cantidad total de items en el sistema.


alejsdev:
Hi @mldaill, thanks for your interest in contributing to the project! However, the existing `read_items` already includes logic to count and return the items with the total count. We prefer to keep the current approach to avoid unnecessary duplication. So, I'll close this one. But thanks again for the effort. :rocket:

PR #1402, opened 2024-10-22
Initial commit with Docker setup

mulsewm:


stevleibelt:
My first look raised a big question, why have you added your `.venv` directory?

PR #1416, opened 2024-11-04
♻️ Update import statements in `login.py` 

Minty-cyber:
- Reorganized and deleted redundant imports


Minty-cyber:
@tiangolo @alejsdev have a look

tiangolo:
Thanks! I don't think this is really necessary so I'll pass on it for now. :coffee:

Issue #400, opened 2021-07-16
What happens when the access token expires


mlmarius:
Hi!

Looking through the sources of the backend and frontend I can't figure out what happens when the JWT access token expires. Is there a strategy in place to extend the access token if the user is using it frequently?

csmcallister:
When the access token expires, a user would need to login again, sending a request to the backend's `/login/access-token` route to get a new access token.

emsi:
@csmcallister: Where in the code it is actually enforced?

csmcallister:
Take a look here first:

https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/490c554e23343eec0736b06e59b2108fdd057fdc/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/api_v1/endpoints/login.py#L36

Then follow the core.security import to here:

https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/490c554e23343eec0736b06e59b2108fdd057fdc/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/core/security.py#L15

As you can see there, the expiration is either pulled in from the app settings or can be overridden with a custom value. The jwt library is responsible for taking in that expiration value. Their docs explain how expiration is determined and enforced: https://pyjwt.readthedocs.io/en/latest/usage.html

maximiliancw:
@emsi stumbled upon this today. In addition to @csmcallister's hints, you can use the `/login/test-token` endpoint (or any other one for that matter) to check if you're current token is still working. It feels a bit hacky though...

hollowdew:
Is there something planned to handle the expiration? 
Like implement a refresh endpoint in the backend and call the endpoint in the frontend?

danieldiamond:
Yeah I'm definitely experiencing and endless loop of `403 Forbidden` requests to `http://localhost/api/v1/users/me` when attempting to access the dashboard with an expired access_token.

It seems that, it doesn't actually check if token is expired, only that it is present. 

https://github.com/fastapi/full-stack-fastapi-template/blob/5005e5a5c17716c8a51bc3fa162e994c1c6035cd/frontend/src/hooks/useAuth.ts#L16-L18

abrichr:
Same issue here. If the user leaves the browser window open for a while then comes back, the app appears to be broken, i.e. the backend returns 403, but the frontend doesn't report this to the user, only hangs.

Suggestions:

[...]

PR #1424, opened 2024-11-16
fix: #400 logout because token expired or corrupted

storenth:
Rename state `error` to get `useQuery`'s native `error` via object destructuring to solve the spinner endless loop in case of backend reboot, expired or corrupted token.


storenth (commented):
Self review done, locally tested by exploratory

YuriiMotov:
@storenth, thanks for your interest and efforts!
Could you please resolve conflicts?

Related discussion (?): https://github.com/fastapi/full-stack-fastapi-template/discussions/1249

PR #1425, opened 2024-11-18
Using TypeAlias to address IDE warnings

javadzarezadeh:
I have added `TypeAlias` typing for dependencies because **VS Code Pylance** issued a warning when using them as type annotations in functions.  
**Warning:** *"Variable not allowed in type expression (Pylance)"*


YuriiMotov:
@javadzarezadeh, could you please check if it's still an issue for you?

I can't reproduce it. Pylance infers type and doesn't argue even without explicit `TypeAlias`

javadzarezadeh:
It seems that the issue doesn't exist on my side anymore, either. You can close the PR.

Thanks for your time, @YuriiMotov.

PR #1428, opened 2024-11-21
Putting `extensions.json` under `.vscode`

ycsh-w:
Hey folks. With this file, we can specify what extensions of VSCode should be used together with this template, and people can easily be prompted in VSCode to install them. I only added Python extension from Microsoft as it's needed to run the debugger. Please let me know what you think, thanks!


YuriiMotov:
@ycsh-w, thanks for your interest and efforts!

There is already #1386. I'll close this PR as #1386 was first

PR #1429, opened 2024-11-22
✨ Add private, local only, API for usage in E2E tests

patrick91:


frontend/tests/utils/privateApi.ts
```diff
@@ -0,0 +1,21 @@
+// @ts-ignore
```
patrick91, line 1: @alejsdev what do you think of having this?

currently without this running the build test would fail. The other option would be to ignore the test files when running build, which maybe is better?

alejsdev: We could ignore the test files while running the build :)

tiangolo (commented):
Nice! A couple of small comments.

tiangolo (approved):
Nice! :rocket: :chicken: :tada:

PR #1431, opened 2024-11-24
🐛 Mount `/backend/app/alembic/versions` as a volume inside container during development

javadzarezadeh:
I've mapped the Docker container's Alembic folder to the local directory, ensuring that any auto-generated migration files created by Alembic are automatically copied to the local development environment.


YuriiMotov (approved):
LGTM

Thanks, @javadzarezadeh !

It's said in the [docs](https://github.com/fastapi/full-stack-fastapi-template/blob/master/backend/README.md#migrations):

> As during local development your app directory is mounted as a volume inside the container, you can also run the migrations with alembic commands inside the container and the migration code will be in your app directory (instead of being only inside the container). So you can add it to your git repository.

But it doesn't work this way in current version. It was broken by #1354

This PR fixes the issue with migrations generated from inside the container not being reflected on the host file system.

**Suggestion:** we should probably update the mentioned paragraph in docs:

> As during local development your `/backend/app/alembic/versions` is mounted as a volume inside the container, you can also run the migrations with alembic commands inside the container and the migration code will be in your app directory (instead of being only inside the container). So you can add it to your git repository.

YuriiMotov (commented):
Thanks, @javadzarezadeh !

There seems to be a problem.. If you run `docker compose watch` and create file in `/backend/app/alembic/versions`, then Docker starts syncing service infinitely:

```
Watch enabled
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 2 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
Syncing service "backend" after 1 changes were detected
```

YuriiMotov:
I think [this suggestion](https://github.com/fastapi/full-stack-fastapi-template/discussions/1370#discussioncomment-10909265) should work: 

> Try excluding the alembic folder from watch to fix the refresh

YuriiMotov:
@javadzarezadeh, would you like to continue working on this?
Or, is anybody up to take this over?

tiangolo:
Thanks for the interest! But as the user inside the container is different than the user outside, the generated files would belong to a different user ID, making them non-editable by the regular user (unless using `sudo`), so I prefer to avoid that.

It was like this at some point and I removed it to avoid this problem precisely. So I'll pass on this, but thanks! :coffee:

PR #1434, opened 2024-11-25
♻️ Remove modify id script in favor of openapi-ts config

patrick91:
This PR removes the modify-id script; this is now handled by a custom config for `openapi-ts`.

The behaviour is slightly different, as we are now only renaming the methods and not the types, which is better in cases where we have operations with the same name; see #1429's `create_user` for an example :)


alejsdev (approved):

patrick91:
I'll merge this once Sebastian reviews the OS PRs 😊

tiangolo (approved):
Looks good! Thank you! :rocket: :tada:

PR #1439, opened 2024-12-02
🎨 Move `prefix` and `tags` to routers

patrick91:


tiangolo (approved):
Nice, thank you! :rocket: :tada:

PR #1449, opened 2024-12-12
👷 Update item edit form data after successful save

flipkick:
Currently, after submitting an item edit form, the item is saved, but the old data is shown when you reopen the form.


alejsdev:
Hello @flipkick, I couldn't replicate the issue you're reporting, it would be better to start with a GitHub discussion. Given that, I'll close this one but thanks for your interest! :rocket:

PR #1450, opened 2024-12-14
🐛 Change get_user_by_email to be case insensetive

grabov:
Fix an issue where `get_user_by_email`  won't return a user object if the requested email case doesn't match the string in the database.

### DIFF
--- a/backend/app/crud.py
+++ b/backend/app/crud.py
@@ -1,7 +1,7 @@
 import uuid
 from typing import Any
 
-from sqlmodel import Session, select
+from sqlmodel import Session, func, select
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks, in this case I think it would be better to lowercase it before passing it to SQLAlchemy/SQLModel, and ensure it is lowercase before saving it to the databse.

Now, in the case of emails, IIRC, domains are case insensitive, but email usernames are not even if many providers consider them case insensitive, it's not part of a standard. So, I would consider this a special use case for a specific app, not something that would apply in general.

Given that, I'll pass on this one, but thanks for the interest! :coffee:

PR #1452, opened 2024-12-20
📝 Update `frontend/README.md` to also remove Playwright when removing Frontend

youben11:
To remove the frontend, we need to remove playwright service as well.


alejsdev:
Thank you! @youben11 :rocket:

PR #1454, opened 2024-12-21
⬆️ Update Dockerfile to use uv version 0.5.11

alejsdev:
⬆️ Update Dockerfile to use uv version 0.5.11


tiangolo (approved):
Nice, thank you! :rocket:

PR #1457, opened 2025-01-10
Migrate to pnpm

acyment:
I simply migrated from npm to pnpm to benefit from increased performance. I had to add zod to package.json because of the way npm and pnpm deal with dependencies. It also made sense to add it as an explicit dependency, as it is being imported in two files in the frontend/src/routes/_layout folder (admin.tsx & items.tsx).

Cheers!

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -5,13 +5,15 @@ WORKDIR /app
 
 COPY package*.json /app/
 
-RUN npm install
+RUN corepack enable
+
+RUN pnpm install
 
 COPY ./ /app/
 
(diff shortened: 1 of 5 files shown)


wesenbergg (commented):
I believe you are missing pnpm-lock.yaml file

tiangolo:
Thanks for the interest! For now, I prefer to keep the template simple with plain npm. I think it's totally fine for people and teams to adopt pnpm if preferred, but for this template I'll keep it as it is for now.

Given that, I'll close this one, but thanks for the effort! :coffee:

PR #1463, opened 2025-01-17
📝 Update development notes

eqsdxr:
Fix incosnistency in capitalization

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -30,6 +30,7 @@
 
 ### Docs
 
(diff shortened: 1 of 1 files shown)


development.md
```diff
@@ -204,4 +204,4 @@ Adminer: http://localhost.tiangolo.com:8080
 
 Traefik UI: http://localhost.tiangolo.com:8090
 
-MailCatcher: http://localhost.tiangolo.com:1080
\ No newline at end of file
+MailCatcher: http://localhost.tiangolo.com:1080
```
wesenbergg, line 207: Diff looks same, why it changed? 🤔

eqsdxr: They are the same; It seems like it's something to do with newline issues

PR #1476, opened 2025-02-01
📝 Add docker-compose version requirement

westonplatter:
If users do not have the min version installed, docker compose watch will fail. See https://docs.docker.com/compose/how-tos/file-watch/


westonplatter:
@alejsdev Thanks for labeling this issue. What can I do on my end of merge this into the docs?

PR #1481, opened 2025-02-04
♻️ Remove redundant item deletion code leveraging cascade delete

nauanbek:
Remove manual item deletion - cascade delete handles this


alejsdev:
Good catch, thank you! @nauanbek :rocket:

PR #1491, opened 2025-02-16
🐛 Fix test failed after test_reset_password if user use own password

andyyoung01:
If user set password of FIRST_SUPERUSER_PASSWORD to a different value of "changethis" in .env file, after run test_reset_password case, the password will be changed to "changethis" becasue it hardcorded in this testcase.
Following testcases will all be failed because of wrong password.
Change it to reset to password in .env file.


alejsdev:
Hi @andyyoung01, this was already handled in #1499 by refactoring the whole test. I'll close this one but thanks for your effort. :raised_hands:

PR #1494, opened 2025-02-17
📝 Removing deprecated manual client SDK step

chandy:
Since openapi-ts was upgraded and  improved(https://github.com/fastapi/full-stack-fastapi-template/pull/1434, https://github.com/fastapi/full-stack-fastapi-template/pull/1430), we no longer need this step to simplify the names since the library will do this for us.


alejsdev:
Good catch, thank you! @chandy :raised_hands: :100:

PR #1499, opened 2025-02-18
🐛 Refactor reset password test to create normal user instead of using super user

alejsdev:
🐛 Refactor reset password test to create normal user instead of using super user


tiangolo (approved):
Looks good, thank you! 🍰

PR #1505, opened 2025-02-20
send account creation email asynchronously via BackgroundTasks

saleemasekrea000:
Making it a background task improves responsiveness and user experience.
For future
If you need to perform heavy background computation and you don't necessarily need it to be run by the same process (for example, you don't need to share memory, variables, etc), you might benefit from using other bigger tools like [Celery](https://docs.celeryq.dev/)
According to the FastAPI docs :
https://fastapi.tiangolo.com/tutorial/background-tasks/?h=#technical-details

### DIFF
--- a/backend/app/api/routes/users.py
+++ b/backend/app/api/routes/users.py
@@ -1,7 +1,7 @@
 import uuid
 from typing import Any
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, BackgroundTasks
 from sqlmodel import col, delete, func, select
 
 from app import crud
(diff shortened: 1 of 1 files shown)


saleemasekrea000:
@alejsdev 
Hi, is there any kind of feedback?
It's been a long time...

YuriiMotov (commented):
@saleemasekrea000, hello!
Thanks for the interest in project!

The problem here is that `BackgroundTasks` can fail silently and in this case user will not receive the email, and admin will not notice it.
With Celery (or other similar tools) we would have retry mechanism to ensure that email has been sent.

Having this I'm not sure it's good idea to move this to BG tasks

jishnu70:
Hi @YuriiMotov , 
This is my first-ever open-source contribution. I’d like to propose making background task support optional:  

- Introduce a boolean `USE_CELERY` in `config.py`.  
- If `USE_CELERY=True`, the developer can start Celery + RabbitMQ via a Docker Compose profile, for example:  
      docker-compose --profile celery up  
- If `USE_CELERY=False`, only the backend runs:  
      docker-compose up  

is this approach acceptable?

tiangolo:
Thanks all! I think it's better to make sure the email is properly sent or scheduled before returning the response, so that if it fails, an error would be returned. So I'll pass on this, but thanks for the interest. :coffee:

PR #1525, opened 2025-03-23
📝 Add an example of command to run the `generate-client.sh` with uv

AllenFang:
Since we use `uv` to manage the py package and virtual env, so suggest to use `uv` to executing client generation for avoiding the packages or python command not found,


tiangolo:
Thanks! I handled this in another PR, including the explicit bash part, and I updated the script to use uv and bun internally, etc.

I'll now close this one. :coffee:

PR #1529, opened 2025-03-26
configure uv to use the latest version

Wetinee:
### Reasons
1. I encountered an issue when building Docker images with the specific uv version (0.5.11). And switching to latest helped resolve the issue. 

1. Using latest ensures that we don’t have to manually update the version in the future. This approach aligns with the recommendations in the uv documentation.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /app/
 
 # Install uv
 # Ref: https://docs.astral.sh/uv/guides/integration/docker/#installing-uv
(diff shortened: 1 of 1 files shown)


stevleibelt:
`uv`  is in high development mode. This could result in hard to reproduce issues when everyone could have a different latest version installed.

On the other hand, even not labeled as version 1.x, I am using uv in production already at it feels pretty mature in the sense of "I haven't encountered many cli/api changes since first use".

I vote for :+1: until `uv` has reached version `1.0.x`.

Thanks for the proposal @Wetinee

tiangolo:
Thanks! This was handled in another PR, so I'll close this one now. :coffee:

PR #1539, opened 2025-04-07
feat: Use bcrypt directly instead of passlib

karta9821:
The passlib library, previously used for password hashing in this project, is no longer actively maintained — its last update was over 4 years ago. This has been officially confirmed here: [link to the issue](https://foss.heptapod.net/python-libs/passlib/-/issues/187).

To improve long-term security and maintainability, this PR replaces passlib with direct usage of the bcrypt library, which is actively maintained and widely used for secure password hashing.


YuriiMotov:
`passlib` was recently replaced with `pwdlib` with Argon2 in FastAPI repository: https://github.com/fastapi/fastapi/pull/13917

So, we should probably be consistent here and also use `pwdlib` in this repo

fudanglp:
@YuriiMotov 
  I prefer this approach over [#1941](https://github.com/fastapi/full-stack-fastapi-template/pull/1941). 

  Compared to replacing passlib with pwdlib:
  - Fewer dependencies: No need for pwdlib, argon2-cffi, cffi, pycparser
  - Simpler code: Direct bcrypt calls vs another password hashing abstraction
  - No unnecessary features: Argon2 support isn't needed if the existing hashes are bcrypt

  The bcrypt library is already a transitive dependency through passlib, so this is essentially just removing the passlib wrapper.

tiangolo:
Thanks! This was handled here: https://github.com/fastapi/full-stack-fastapi-template/pull/2104

I'll close this one now. :coffee:

PR #1549, opened 2025-04-16
exploration/dps/ Include adminer to clean template using Sqlite

dpsancha:


.env
stevleibelt, line 1: Please remove the env file from your commit.

Especially the change of the PROJECT_NAME or the STACK_NAME is a clear indicator that you've made a mistake here.

stevleibelt:
I've stopped reviewing your merge request.

What is the big benefit behind it?

Could you please explain what you've done and why this is a benefit for all of the users who want to have a generic or general template to build up their project from this repository?

PR #1556, opened 2025-04-19
add health check point

mohamedachrefhacheni:
This PR introduces a health check endpoint to monitor service availability and database connectivity.
Impact:
Enables better monitoring
Helps DevOps with uptime checks
Follows REST API best practices

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -20,6 +20,11 @@ app = FastAPI(
     generate_unique_id_function=custom_generate_unique_id,
 )
 
+# Add health check endpoint directly to main app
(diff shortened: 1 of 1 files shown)


YuriiMotov (changes requested):
@mohamedachrefhacheni, thanks for your interest and efforts!

We already have health check endpoint: https://github.com/fastapi/full-stack-fastapi-template/blob/8af907c763c85c38a2dc745792eda741de5384bb/backend/app/api/routes/utils.py#L29-L31
So, changes should be made there instead of adding one more.

Also, all changes should be covered by tests

PR #1563, opened 2025-04-24
Implement search filter functionality in backend

Durganjali-sidda:
Implemented search filter functionality in the backend for the /items route. The endpoint now supports dynamic search with filtering options for title and keyword across both title and description. Additionally, the filtering respects user permissions, ensuring non-admin users can only access their own items, while superusers can access items from other users. This enhancement improves item search and retrieval, providing more flexible querying for both regular and administrative users.

Key updates include:

Added query parameters title, keyword, and owner_id for filtering items.

Improved permission handling for item access based on user roles.

Enhanced item retrieval with support for filtering based on title and description.

### DIFF
--- a/backend/app/api/routes/items.py
+++ b/backend/app/api/routes/items.py
@@ -12,31 +12,35 @@ router = APIRouter(prefix="/items", tags=["items"])
 
 @router.get("/", response_model=ItemsPublic)
 def read_items(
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest. This would be better as a feature of your own project instead of here. So I'll pass on this one. :coffee:

PR #1567, opened 2025-04-25
Update reset password message

azimjonn:
Remove the ending period in the password reset success message to pass playwright tests.


alejsdev:
Hi @azimjonn, thanks for your interest, but I don't think this is necessary since tests are passing. Given that, I'll close this PR, but thanks for your effort! :rocket:

PR #1573, opened 2025-04-27
👷🏻‍♀️  Update CI for client generation

alejsdev:
👷🏻‍♀️ Update CI for client generation 

Streamline development to run playwright tests locally.


tiangolo (approved):
Looks good, thanks! :rocket: 

As the private client in the frontend would be removed by tree shaking during the JS build, it's fine to have it in git, and the developer experience for developers writing Playwright tests would be better, so yep, agreed this works.

PR #1574, opened 2025-04-27
Exploration/dps/rename helm

dpsancha:


nnWhisperer:
Hello @dpsancha , this is an important feature, why did you close it?
I checked that the helm chart doesn't have traefik's labels, which is an important part to make it work. May be I oversighted something. Thanks for your work

PR #1593, opened 2025-04-29
Enhance Database Modeling and SQLAlchemy Compatibility

cubxxw:
This pull request makes several changes to the `backend/app/models.py` file to enhance database modeling and improve compatibility with SQLAlchemy. The key updates include adjustments to field definitions, changes to relationships, and the removal of certain constraints.

### Changes to field definitions:

* Updated the `email` field in `UserBase` to explicitly use `sa_type=String(255)` for better SQLAlchemy compatibility.
* Changed the `email` field type in `UserRegister`, `UserUpdate`, and `UserUpdateMe` from `EmailStr` to `str` for consistency and to avoid type-related issues.

### Changes to relationships:

* Modified the `items` relationship in the `User` model to include `sa_relationship_kwargs` for defining cascading behavior and specifying join conditions.
* Updated the `owner_id` field in the `Item` model to remove the `foreign_key` constraint while keeping it indexed for performance. Added `sa_relationship_kwargs` to the `owner` relationship for explicit join conditions.

### DIFF
--- a/backend/app/models.py
+++ b/backend/app/models.py
@@ -2,11 +2,14 @@ import uuid
 
 from pydantic import EmailStr
 from sqlmodel import Field, Relationship, SQLModel
+from sqlalchemy import String
 
 
 # Shared properties
 class UserBase(SQLModel):
-    email: EmailStr = Field(unique=True, index=True, max_length=255)
+    email: EmailStr = Field(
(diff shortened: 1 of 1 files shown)


tiangolo:
Yep, I think we don't currently need this, so I'll pass on this for now, but thanks for the interest! :coffee:

PR #1624, opened 2025-05-12
♻️ Changing all status codes from raw numbers to using constants from `status`

javadzarezadeh:
I believe that using status codes via FastAPI (Starlette, actually!) instead of using raw numbers is more readable, consistent, and makes development easier.

Re-request from PR #1319.


javadzarezadeh:
Dear @alejsdev 
Please add the 'refactor' label. Thanks!

javadzarezadeh:
Thank you, @YuriiMotov.
I've made the changes you requested. I even changed the numbers in ranges to starlette HTTP codes. If you still think that numbers are better for ranges. Let me know. I'll revert it quickly.

YuriiMotov (approved):
LGTM

Still not sure about ranges.
I think we should leave numbers for 
```py
assert 200 <= r.status_code < 300
```
But these changes are in [separate commit](https://github.com/fastapi/full-stack-fastapi-template/pull/1624/commits/2d800ed7903686ca62a4d6beafc4db6e257fc158). Can be easily reverted

@javadzarezadeh, thanks again!

tiangolo:
Thanks! Actually, recently we had to move some of them that were using the names from `status` to use the numbers, because the name was changed in a recent version of Starlette. So, for now, I'll pass on this one, but thanks! :coffee:

PR #1542, opened 2025-04-13
Backend update

marcelomizuno:
Criadas as endpoints para gerenciamento dos tickets de atendimento.

PR #1628, opened 2025-05-17
Frontend update

marcelomizuno:


stevleibelt:
@marcelomizuno you really should be more careful with your pull requests.

This fits into your [backend update](https://github.com/fastapi/full-stack-fastapi-template/pull/1542) pull request.
You've pushed a lot of private information and code changes that look more like things you need but that should not be in a generic template repository.

Issue #125, opened 2020-03-01
Separating database for tests and dev


robido:
Each time the tests are run, a few random users are created. I can see that a problem when I want to test models, and then the database becomes cluttered really quickly, even if just in dev.

To have two databases, I tried to split env-postgres.env in two files:
     env-postgres.env
     env-postgres-test.env

Then in docker-compose.test.yml I changed the env_file entry to refer to the new env file.

After that, I run `docker-compose up -d`

When I try to run the tests, I get an error saying the database does not exist. If I create it manually it works, but more errors are showing because the migrations didn't run automatically. How is the database for the backend automatically created, and how can I do the same for the tests database?

Thanks

drone-ah:
It looks like some of the tests access the API from the other container and does black box testing. I am not sure of a simple way to change the database to test these.

Some of the other tests (crud tests etc.) however, can be changed fairly easily to use a separate db.

What I did to achieve this is to add the following two fixtures in `tests/conftest.py`

```python
from sqlalchemy import create_engine
from sqlalchemy.orm import scoped_session, sessionmaker

from app.db import session

from app.db.base_class import Base

from app.db import init_db
 
engine = None

@pytest.fixture(autouse=True, scope='session')
def database_session_fixture():
    global engine
    print("Before Session")

    conn = session.engine.connect()
    conn.execute('commit')
    conn.execute('create database app_automated_tests')

    sqlalchemy_database_uri = (
        f"postgresql://{config.POSTGRES_USER}:{config.POSTGRES_PASSWORD}@{config.POSTGRES_SERVER}/app_automated_tests"
    )

    engine = create_engine(sqlalchemy_database_uri, pool_pre_ping=True)
    session.db_session = scoped_session(
        sessionmaker(autocommit=False, autoflush=False, bind=engine)
    )
    session.Session = session.db_session

    yield

    print('after session')
    engine.dispose()
    conn.execute('commit')
    conn.execute("drop database if exists app_automated_tests")


@pytest.fixture(scope="function")
def db_session():
    Base.metadata.create_all(engine)
    sess = session.Session()
    print("INit DB")
    init_db.init_db(sess)
    print("Yield")
    yield sess
    sess.close()
    Base.metadata.drop_all(bind=engine)
```

I then updated the crud tests, for example in test_item, I removed the following import
`from app.db.session import db_session`

and updated each of the tests to take a parameter of `db_session`

[...]

PR #1629, opened 2025-05-18
feat(config): add support for separate test database via .env.test

EmiFunes91:
### Summary

This PR introduces support for using a separate PostgreSQL database during testing by adding a new `.env.test` file and dynamic configuration logic.

### Changes

- Adds `.env.test` to define `ENVIRONMENT=test` and `POSTGRES_DB=app_test`
- Updates `config.py` to automatically switch the database based on environment
- Documents usage in `README.md` under a new **Testing Environment** section

### Motivation

Currently, the development and testing environments use the same database (`app`), which can lead to data collisions or accidental data loss.

This PR addresses [Issue #125](https://github.com/fastapi/full-stack-fastapi-template/issues/125) by separating configuration and runtime behavior cleanly, without introducing breaking changes.

### Notes

- Docker Compose is reused for both dev and test; no new service is introduced
- This is backwards compatible and does not interfere with existing setups
- CI environments can now easily isolate their test database

Let me know if you'd like `.env.test` to be excluded via `.gitignore` or if you'd prefer a `docker-compose.override.yml` for full isolation.


EmiFunes91:
Hi maintainers!

The `check-labels` workflow is currently failing because it requires one of the following labels:
`breaking`, `security`, `feature`, `bug`, `refactor`, `upgrade`, `docs`, `lang-all`, or `internal`.

Could you please add one of them? I believe `feature` would be appropriate in this case.

Thank you very much!

YuriiMotov:
@EmiFunes91, have you checked whether it really works (separate db is used)?
I haven't run it yet, but it seems that the implementation is incomplete and it will not work

PR #1630, opened 2025-05-18
Modular-monolith-refactoring

Fguedes90:
- Updated SMTP configuration in .env and docker-compose.override.yml to use MailHog for email testing
- Added utils module with health check and test email endpoints
- Implemented event handling for password reset requests, integrating with the email service
- Improved email service to send password reset emails upon event trigger
- Removed deprecated files and updated documentation to reflect new module structure

These changes contribute to the ongoing development of the modular monolith architecture, enhancing email functionality and introducing utility endpoints for better service management.


stevleibelt:
@Fguedes90,

thank for your work.

You've changed the obvious `changethis` to a more regular looking secret key or fist superuser password in the configuration. Is there a reason for this change?

Why have you removed the whole Copier-Section in the readme?

You've added a TASKS.md with a lot of unfinished tasks. This looks like your change is not fully done.

Is there a reason why you've replaced the code in `get_db` with just another method call?

Inside the `get_module_settings`, would it make sense to replace the `module_name: str` with a `module_name: ModuleNameStrEnum`? Creating modules but have a central module configuration file I need to adapt does not look like decoupled software, but that is just my gut-feeling. Returning an empty setting instead of raising an exception looks also like a well introduced code area where you can spend hours for debugging.

I must admint, I've stopped looking over your code after my scrollbar was at 33 percent of the available slide position. This is a really big change, so my points mentioned above are just the first thoughts.

PR #1672, opened 2025-06-15
🐛 Fix `parse_cors` function to be consistent for both empty string and empty list

rolkotaki:
Fix the `parse_cors` function in `config.py` so that it behaves consistently for both empty string and empty list.
Currently, when an empty list is provided as an argument, it returns an empty list (`[]`). However, when an empty string (`""`) is provided, it returns a list with an empty string (`[""]`) and starting the FastAPI application fails.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -18,6 +18,8 @@ from typing_extensions import Self
 
 def parse_cors(v: Any) -> list[str] | str:
     if isinstance(v, str) and not v.startswith("["):
+        if not v.strip():
+            return []
         return [i.strip() for i in v.split(",")]
     elif isinstance(v, list | str):
         return v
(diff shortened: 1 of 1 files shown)


rolkotaki:
Anyone who is a maintainer, could you please add a label to this PR? Thanks.

backend/app/core/config.py
```diff
@@ -18,6 +18,8 @@
 
 def parse_cors(v: Any) -> list[str] | str:
     if isinstance(v, str) and not v.startswith("["):
+        if not v.strip():
+            return []
         return [i.strip() for i in v.split(",")]
```
YuriiMotov, line 23: How about:
```py
        return [i.strip() for i in v.split(",") if i.strip()]
```

YuriiMotov: With this update we don't need `if not v.strip():` anymore

YuriiMotov (approved):
LGTM

Thanks!

tiangolo (approved):
Cool, thanks! 🚀

Issue #383, opened 2021-06-08
include files with password in .gitignore


erdnax123:
for security reasons:

create .env.local and add to .gitignore, move all config with passwords to .env.local

add cookiecutter-config-file.yml to .gitignore

tiangolo:
Yep, you can definitely do that! Nevertheless, that would be directly on your repo, not in this one that is the template with the defaults to get something runnable right after cloning. Thanks for the interest! :coffee:

Issue #218, opened 2020-06-20
Ignore .env by default


paxcodes:
Thanks @tiangolo for another awesome tool.

I was wondering if we should add the .env in the .gitignore file by default. 

I understand people have different workflows but I wonder if it would be better that the .env file is _explicity_ exposed if the workflow demands it, instead of needing to explicitly hide the .env file which usually contain secrets.

paxcodes:
Looks like it's been discussed already. I should have searched more closely.

https://github.com/tiangolo/full-stack-fastapi-postgresql/issues/122
https://github.com/tiangolo/full-stack-fastapi-postgresql/pull/128

tiangolo:
Yep, thanks for closing the issue :+1:

PR #1684, opened 2025-06-19
📝 Add additional notes about `.env` and `.copier-answers.yml` not being in `.gitignore` by default

tylermilner:
Based on previous discussions (#724, #1119, #383, #218, #128, #122), it's clear that we don't want to go ahead and add files containing secrets (e.g. `.env`, `.copier-answers.yml`) to the template's `.gitignore` by default.

This PR surfaces that information to the main `README.md`. Hopefully, this will help make that decision more visible to those getting started with the project and reduce additional issues/discussions opened on the topic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ You can (and should) pass these as environment variables from secrets.
 
 Read the [deployment.md](./deployment.md) docs for more details.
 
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest, but I'll pass on this one for now. :coffee:

PR #1688, opened 2025-06-21
🐛 Fix loginAccessToken Prop in useAuth.ts to Resolve 400 Error on dashboard login

gregorvolkmann:
This pull request fixes an HTTP 400 error in production by updating the login function in frontend/src/hooks/useAuth.ts. The LoginService.loginAccessToken call was incorrectly passing { formData: data } instead of { formData: { data.username, data.password } }, causing a type mismatch and sending multipart/form-data instead of application/x-www-form-urlencoded. The change ensures compliance with the LoginLoginAccessTokenData type and corrects the login request format.

### DIFF
--- a/frontend/src/hooks/useAuth.ts
+++ b/frontend/src/hooks/useAuth.ts
@@ -46,7 +46,8 @@ const useAuth = () => {
       formData: {
         username: data.username,
         password: data.password,
-      })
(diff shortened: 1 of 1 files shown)


gregorvolkmann:
Yes. No changes are needed.

PR #1715, opened 2025-07-13
Add search functionality to items endpoint with tests

yaswanthrajeev:
### Summary

- Introduced a `search` query parameter to the `/api/v1/items/` endpoint.
- Enables searching items by title or description (case-insensitive, partial match).

### How to Test

```bash
curl -X GET "http://localhost:8000/api/v1/items/?search=example" \
  -H "Authorization: Bearer <your_access_token>"
```
```

It will render with headings, lists, and code blocks.

---

**You’re good to go—just copy and paste!**

### DIFF
--- a/backend/app/api/routes/items.py
+++ b/backend/app/api/routes/items.py
@@ -25,7 +25,6 @@ def read_items(
     if current_user.is_superuser:
         query = select(Item)
         count_statement = select(func.count()).select_from(Item)
-        statement = select(Item).offset(skip).limit(limit)
     else:
         query = select(Item).where(Item.owner_id == current_user.id)
         count_statement = (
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by YuriiMotov** [comment]
@yaswanthrajeev, thank you for the interest in project!

Do you have plans to continue working on this?
For now it looks incomplete. I'll convert it to draft. Press "Ready for review" when it's ready

File: backend/app/tests/api/routes/test_items.py [hunk 1]
```diff
@@ -162,3 +162,102 @@ def test_delete_item_not_enough_permissions(
...
+    
+    # Test search with pagination
+    response = client.get(
+        f"{settings.API_V1_STR}/items/?search=Test&skip=0&limit=2",
+        headers=superuser_token_headers,
+    )
+    assert response.status_code == 200
+    content = response.json()
+    assert content["count"] == 2
+    assert len(content["data"]) == 2
```
youneshenniwrites, line 263: Maybe also assert the titles of the returned items (e.g., "Test Item 1" and "Test Item 2") so the test is stronger and doesn’t rely on implicit ordering.

youneshenniwrites (changes requested):

PR #1721, opened 2025-07-18
🚨 Fix correct return type for `SQLALCHEMY_DATABASE_URI` computed field

myrkytyn:
Replace MultiHostUrl.build() with PostgresDsn.build() to match the PostgresDsn return type annotation and resolve type checker error

Fix for the error:
> Type "MultiHostUrl" is not assignable to return type "PostgresDsn"
>  "MultiHostUrl" is not assignable to "PostgresDsn"

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -11,7 +11,6 @@ from pydantic import (
     computed_field,
     model_validator,
 )
-from pydantic_core import MultiHostUrl
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from typing_extensions import Self
 


YuriiMotov:
It was initially `PostgresDsn`, but changed to `MultiHostUrl` in this commit: https://github.com/fastapi/full-stack-fastapi-template/pull/667/commits/436952403a8ff6534d463065a321ead434c17f62
There is no explanation why..

As an alternative, we may want to change the return type to `MultiHostUrl` to make it more general.

BTW, I couldn't reproduce the type checker error. What type checker do you use?

PR #1722, opened 2025-07-21
feat(i18n): Add multi-language support to realize Chinese and English switching function

frontend-winter:
## Description

This pull request adds multi-language (i18n) support to the frontend, enabling users to switch between Chinese and English.

## Changes

- Integrated i18n support using [i18next/vue-i18n/...]
- Added language switcher to the header
- Extracted all static texts in the frontend to translation files
- Default language is English; user can toggle to Chinese
- Language preference stored in localStorage (or cookies/session)

## Motivation

As the project aims to be internationally usable, adding multilingual support improves user experience and accessibility for non-English speakers.

## Related Issues

N/A

## Screenshots

image
image

## Checklist

- [x] All strings externalized into translation files
- [x] Tested language switch works properly
- [x] Fallback to English if translation missing

Hi maintainers 👋

This PR adds i18n support to enable Chinese/English language switching in the frontend.

Could you please help add one of the required labels (e.g. `feature` or `lang-all`) so that the label-checker CI can pass? 🙏 Thank you!

---

该 PR 实现了前端中英文语言切换功能。

CI 校验失败是因为没有打标签（例如 `feature` 或 `lang-all`），麻烦协助加一个标签使其通过，非常感谢 🙏


tiangolo:
Thanks! This would be better as a feature of your own project and not something that comes here by default. So, I'll pass on this one. :coffee:

frontend-winter:
Thanks for the quick reply @tiangolo ☕ I totally understand your concern about keeping the core minimal.

Since internationalization could still be useful for some community members, would it make sense if I move this work into a separate repo/template (e.g. `fastapi-i18n-template`) and add it to the ecosystem/resources list?  

That way, people who need multi-language support can directly use it without adding extra burden to the main repo.  
I’m happy to maintain and document it on my side. 🚀

What do you think?

PR #1763, opened 2025-08-11
🚸 Add Postgres password min length validation in `copier.yml`

angelmoya:
Ensure min lenght for postgresql password

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -55,7 +55,7 @@ postgres_password:
   default: changethis
 validate: |
   {% if postgres_password|length < 8 %}
-  Password must have at least 8 characters
+  Password must have at least 8 characters. Try again
   {% endif %}
 
 sentry_dsn:


YuriiMotov (dismissed):
LGTM

The idea of this PR is that PostgreSQL has a min password length constraint, and it can be confusing for people who don't know this - they can use password shorter than 8 characters and then the stack will not run.

```
service "prestart" didn't complete successfully: exit 1 
```

---

With changes from this PR applied, copier will show a helpful message and ask to try again:

```
...
🎤 'The password for the PostgreSQL database, stored in .env,
you can generate one with:
python -c "import secrets; print(secrets.token_urlsafe(32))"'
   ch
🎤 validate
 
Password must have at least 8 characters. Try again
 
changethis
🎤 The DSN for Sentry, if you are using it, you can set it later in .env
...

```

---

To check, use the command below:
```
copier copy --vcs-ref patch-1 https://github.com/angelmoya/full-stack-fastapi-template.git my_project --trust
```

YuriiMotov:
@angelmoya, this validation works, but I can't get what problem does it solve?
I thought that PostgreSQL has min password length constraint and it can be tricky to find why the Stack doesn't run. But it turned out that DB can run even with short password.

Could you please clarify the purpose of this PR?

Issue #1780, opened 2025-08-20
Broken unit tests exposed by Python 3.12+


vicaya:
The 2 `test_init_successful_connection` tests in `backend/app/tests/scripts/test_\*_pre_start.py` are basically noops, due to misuses of MagicMock(return_value=True). The 2 unit tests would error out due to unknown attributes (`called_once_with`) when running `bash scripts/test.sh` under python 3.12+, which tightened up the MagicMock implementation.

PR would follow.

tiangolo:
Thanks for the PR! 🚀

PR #1781, opened 2025-08-20
✅ Fix incorrect mocking in unit tests (issue #1780)

vicaya:
Fix issue #1780: the 2 `test_init_successful_connection` tests were basically noops due to misuse of MagicMock.

The errors went undetected until python 3.12+ tightened up the MagicMock implementation.


YuriiMotov (approved):
LGTM

I suggest removing several lines to simplify tests (I tested it locally - it doesn't break tests).

Currently if you replace `session.exec(select(1))` in [this line](https://github.com/fastapi/full-stack-fastapi-template/blob/8af907c763c85c38a2dc745792eda741de5384bb/backend/app/backend_pre_start.py#L26) with `pass`, [this assertion](https://github.com/fastapi/full-stack-fastapi-template/blob/8af907c763c85c38a2dc745792eda741de5384bb/backend/app/tests/scripts/test_backend_pre_start.py#L31) will not fail as it's expected.
This PR fixes mocking the `Session` and after applying it, that test will not only ensure that there was no exception in [init](https://github.com/fastapi/full-stack-fastapi-template/blob/8af907c763c85c38a2dc745792eda741de5384bb/backend/app/backend_pre_start.py#L22C5-L22C9), but also that `select(1)` was actually called.

The same for https://github.com/fastapi/full-stack-fastapi-template/blob/master/backend/app/tests/scripts/test_test_pre_start.py

@vicaya, thank you!
~~Could you please take a look at my suggested changes?~~ (done)

tiangolo:
It seems we now have some conflicts here :thinking: :pensive:

PR #1799, opened 2025-08-29
BUG: fix system python dependency

alexrockhill:
From testing, this was actually using the system-wide python install so probably should set up its own for a better config.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -124,6 +124,10 @@ jobs:
       - test-playwright
     runs-on: ubuntu-latest
     steps:
+      - name: Install python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@release/v1
         with:


YuriiMotov:
@alexrockhill, thanks for interest in this project!

The `re-actors/alls-green` cation only has standard dependencies and doesn't have any specific requirements on Python version.
Also, current implementation is inline with the [official example](https://github.com/re-actors/alls-green/blob/release/v1/README.md#usage).

Please, explain why you think this is important.
Installing Python will increase execution time. Do we need this?

alexrockhill:
I ran this on a local runner without a system wide python installation and it failed because all-greens calls a plain python command within it.

YuriiMotov:
> I ran this on a local runner without a system wide python installation and it failed because all-greens calls a plain python command within it.

Then I would make this step conditional (if no Python 3 installed - install)

stevleibelt:
@YuriiMotov how important is execution/build time compared to "comes with all batteries included"?

The less external and uncontrollable dependencies, the better and streamlined the result, or?

alejsdev:
Hi @alexrockhill :wave: 

This job is to determine if a PR can be merged, and since we don't use local runners, we don't really need this change here, but feel free to add it to your own project. Given that, I'll pass on this one. But thanks for the effort! :rocket:

PR #1833, opened 2025-09-03
BUG: Fix frontend linting

alexrockhill:
Right now, on `master`, `cd frontend; npm run lint` hangs and gives a warning:

```
$ npm run lint

> frontend@0.0.0 lint
> biome check --apply-unsafe --no-errors-on-unmatched --files-ignore-unknown=true ./

internalError/fs  DEPRECATED  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ The argument --apply-unsafe is deprecated, it will be removed in the next major release. Use --write --unsafe instead.
```

This is because it is trying to lint `dist`. If you fix that, it errors out on the generated backend client:

```
$ npm run lint

> frontend@0.0.0 lint
> biome check --apply-unsafe --no-errors-on-unmatched --files-ignore-unknown=true ./

internalError/fs  DEPRECATED  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ The argument --apply-unsafe is deprecated, it will be removed in the next major release. Use --write --unsafe instead.
  

./src/client/core/CancelablePromise.ts:87:10 lint/suspicious/noThenProperty ━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Do not add then to a class.
  
    85 │   }
    86 │ 
  > 87 │   public then<TResult1 = T, TResult2 = never>(
       │          ^^^^
    88 │     onFulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | null,
    89 │     onRejected?: ((reason: unknown) => TResult2 | PromiseLike<TResult2>) | null,
  

./src/client/sdk.gen.ts:51:8 lint/complexity/noStaticOnlyClass ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid classes that contain only static members.
  
     49 │ } from "./types.gen"
     50 │ 
   > 51 │ export class ItemsService {
        │        ^^^^^^^^^^^^^^^^^^^^
   > 52 │   /**
         ...
  > 169 │   }
  > 170 │ }
        │ ^
    171 │ 
    172 │ export class LoginService {
  
  ℹ Prefer using simple functions instead of classes with only static members.
  

./src/client/sdk.gen.ts:172:8 lint/complexity/noStaticOnlyClass ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid classes that contain only static members.
  
    170 │ }
    171 │ 
  > 172 │ export class LoginService {
        │        ^^^^^^^^^^^^^^^^^^^^
  > 173 │   /**
         ...
  > 274 │   }
  > 275 │ }
        │ ^
    276 │ 
    277 │ export class PrivateService {
  
  ℹ Prefer using simple functions instead of classes with only static members.
  

./src/client/sdk.gen.ts:277:8 lint/complexity/noStaticOnlyClass ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid classes that contain only static members.
  
    275 │ }
    276 │ 
  > 277 │ export class PrivateService {
        │        ^^^^^^^^^^^^^^^^^^^^^^
  > 278 │   /**
         ...
  > 298 │   }
  > 299 │ }
        │ ^
    300 │ 
    301 │ export class UsersService {
  
  ℹ Prefer using simple functions instead of classes with only static members.
  

./src/client/sdk.gen.ts:301:8 lint/complexity/noStaticOnlyClass ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid classes that contain only static members.
  
    299 │ }
    300 │ 
  > 301 │ export class UsersService {
        │        ^^^^^^^^^^^^^^^^^^^^
  > 302 │   /**
         ...
  > 511 │   }
  > 512 │ }
        │ ^
    513 │ 
    514 │ export class UtilsService {
  
  ℹ Prefer using simple functions instead of classes with only static members.
  

./src/client/sdk.gen.ts:514:8 lint/complexity/noStaticOnlyClass ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid classes that contain only static members.
  
    512 │ }
    513 │ 
  > 514 │ export class UtilsService {
        │        ^^^^^^^^^^^^^^^^^^^^
  > 515 │   /**
         ...
  > 547 │     })
  > 548 │   }
  > 549 │ }
        │ ^
    550 │ 
  
  ℹ Prefer using simple functions instead of classes with only static members.
  

./src/client/core/request.ts:60:7 lint/complexity/noForEach ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Prefer for...of instead of forEach.
  
    58 │       append(key, value.toISOString())
    59 │     } else if (Array.isArray(value)) {
  > 60 │       value.forEach((v) => encodePair(key, v))
       │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    61 │     } else if (typeof value === "object") {
    62 │       Object.entries(value).forEach(([k, v]) => encodePair(`${key}[${k}]`, v))
  
  ℹ forEach may lead to performance issues when working with large arrays. When combined with functions like filter or map, this causes multiple iterations over the same type.
  

./src/client/core/request.ts:62:7 lint/complexity/noForEach ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Prefer for...of instead of forEach.
  
    60 │       value.forEach((v) => encodePair(key, v))
    61 │     } else if (typeof value === "object") {
  > 62 │       Object.entries(value).forEach(([k, v]) => encodePair(`${key}[${k}]`, v))
       │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    63 │     } else {
    64 │       append(key, value)
  
  ℹ forEach may lead to performance issues when working with large arrays. When combined with functions like filter or map, this causes multiple iterations over the same type.
  

./src/client/core/request.ts:68:3 lint/complexity/noForEach ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Prefer for...of instead of forEach.
  
    66 │   }
    67 │ 
  > 68 │   Object.entries(params).forEach(([key, value]) => encodePair(key, value))
       │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    69 │ 
    70 │   return qs.length ? `?${qs.join("&")}` : ""
  
  ℹ forEach may lead to performance issues when working with large arrays. When combined with functions like filter or map, this causes multiple iterations over the same type.
  

./src/client/core/request.ts:79:25 lint/suspicious/noPrototypeBuiltins ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Do not access Object.prototype method 'hasOwnProperty' from target object.
  
    77 │     .replace("{api-version}", config.VERSION)
    78 │     .replace(/{(.*?)}/g, (substring: string, group: string) => {
  > 79 │       if (options.path?.hasOwnProperty(group)) {
       │                         ^^^^^^^^^^^^^^
    80 │         return encoder(String(options.path[group]))
    81 │       }
  
  ℹ It's recommended using Object.hasOwn() instead of using Object.hasOwnProperty().
  
  ℹ See MDN web docs for more details.
  

./src/client/core/request.ts:103:5 lint/complexity/noForEach ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Prefer for...of instead of forEach.
  
    101 │     }
    102 │ 
  > 103 │     Object.entries(options.formData)
        │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 104 │       .filter(([, value]) => value !== undefined && value !== null)
         ...
  > 110 │         }
  > 111 │       })
        │       ^^
    112 │ 
    113 │     return formData
  
  ℹ forEach may lead to performance issues when working with large arrays. When combined with functions like filter or map, this causes multiple iterations over the same type.
  

./src/client/core/request.ts:107:11 lint/complexity/noForEach ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Prefer for...of instead of forEach.
  
    105 │       .forEach(([key, value]) => {
    106 │         if (Array.isArray(value)) {
  > 107 │           value.forEach((v) => process(key, v))
        │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    108 │         } else {
    109 │           process(key, value)
  
  ℹ forEach may lead to performance issues when working with large arrays. When combined with functions like filter or map, this causes multiple iterations over the same type.
  

./src/client/core/request.ts:153:9 lint/performance/noAccumulatingSpread ━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Avoid the use of spread (`...`) syntax on accumulators.
  
    151 │     .reduce(
    152 │       (headers, [key, value]) => ({
  > 153 │         ...headers,
        │         ^^^^^^^^^^
    154 │         [key]: String(value),
    155 │       }),
  
  ℹ Spread syntax should be avoided on accumulators (like those in `.reduce`) because it causes a time complexity of `O(n^2)`.
  
  ℹ Consider methods such as .splice or .push instead.
  

Checked 81 files in 106ms. No fixes applied.
Found 13 errors.
check ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Some errors were emitted while running checks.
```

The fundamental issue is that ``openapi-ts`` is not and probably reasonably will not always be up to date with the latest biome formatting. (I was able to fix these errors pretty quickly but since they're generated it's not helpful but I can back-contribute up to ``openapi-ts``). I looked into it but unfortunately because there's only one biome.json file, it's not possible to call format in generating the client and call `npm run lint` that ignores that formatted client code. So, sadly, I think you need to ignore it and not lint the generated code. It's already pretty nice though. This is what I would do, there's other approaches but I think it's an issue that `npm run lint` doesn't work and this is one option to fix it.

Also #1098 did not actually work, I'm not sure why.


alejsdev:
Hi @alexrockhill, Biome also needed to be updated to the latest version, so in the latest upgrade #1861 (which also required extra changes), I included this fix proposed here as well. Thank you for reporting! :rocket:

PR #1865, opened 2025-09-09
🔧 Add T201 rule to ruff linting configuration to disallow print statements

alejsdev:
🔧 Add T201 rule to ruff linting configuration to disallow print statements


tiangolo (approved):
🚀🎉

PR #1447, opened 2024-12-11
Update password max length

michaelAlvarino:
On my machine, the command provided to generate a password (using
python3 specifically), results in a 44 character string, while the
maximum allowable character length in the model is 40, causing a
conflict. This PR updates the max_legnth for all password model
definitions.

```
MacBook-Pro-3:full-stack-fastapi-template michaelalvarino$ python3 -c "import secrets; print(secrets.token_urlsafe(32))" | wc -c
      44
MacBook-Pro-3:full-stack-fastapi-template michaelalvarino$ python3 -c "import secrets; print(secrets.token_urlsafe(32))" | wc -c
      44
MacBook-Pro-3:full-stack-fastapi-template michaelalvarino$ python3 -c "import secrets; print(secrets.token_urlsafe(32))" | wc -c
      44
```

PR #1914, opened 2025-10-02
📝 Update code example to limit the length of generated secret value by 40 bytes

YuriiMotov:
In docs we instruct users to use `python -c "import secrets; print(secrets.token_urlsafe(32))"` to generate secrets for all parameters that have default `changethis` value ([source](https://github.com/fastapi/full-stack-fastapi-template/blob/master/README.md#generate-secret-keys)).

Since `token_urlsafe` uses Base64 encoding, the length of generated token is actually not always less than 40 characters (it's 32*1.3=41.6 in average, see [docs](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe)).
This leads to app failure on start due to violating the constraint of first user password (currently limited by 40, there is a PR https://github.com/fastapi/full-stack-fastapi-template/pull/1447 to increase the limit).

I updated the code example to generate secret value in docs and in `copier.yml` (for the sake of consistency).

Alternatively (IMO, even better) we can just merge https://github.com/fastapi/full-stack-fastapi-template/pull/1447

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Some environment variables in the `.env` file have a default value of `changethi
 You have to change them with a secret key, to generate secret keys you can run the following command:
 
 ```bash
-python -c "import secrets; print(secrets.token_urlsafe(32))"
+python -c "import secrets; print(secrets.token_urlsafe(32)[:40])"
 ```
 
 Copy the content and use that as password / secret key. And run that again to generate another secure key.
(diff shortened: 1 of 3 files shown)


svlandeg:
Closing this one, as https://github.com/fastapi/full-stack-fastapi-template/pull/1447 was merged that updates the max length to 128 🎉

PR #1940, opened 2025-10-18
🐛 Fix `EMAILS_FROM_NAME` type to be `str` instead of `EmailStr`

martin0258:
`EMAILS_FROM_NAME` has default `PROJECT_NAME`, which is a string like "Full Stack FastAPI Project" and not a valid email address. Therefore, the type of `EMAILS_FROM_NAME` should be changed from email to str.


YuriiMotov (approved):
LGTM!

@martin0258, thanks!

alejsdev:
Good catch, thank you! @martin0258 :tada:

PR #1971, opened 2025-11-07
Adding async backend example that mirrors the current synchronous backend

A00474880:
I think having an asynchronous example here would be helpful.

Please let me know if this is good and I'm happy to make changes and help maintain.


A00474880:
Hi there, I'd really love to see this gets merged. Can I get some feedback on this?

PR #2025, opened 2025-12-03
fix: Ensure consistent passphrase length

jakob1379:
This PR ensures that the generated passphrases have a consistent length. 

During the docker compose, some service complained that the secret were too long, which is caused by the `python -c "import secrets; print(secrets.token_urlsafe(32))` which is in bytes, not characters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Some environment variables in the `.env` file have a default value of `changethi
 You have to change them with a secret key, to generate secret keys you can run the following command:
 
 ```bash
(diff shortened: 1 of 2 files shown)


tiangolo:
Which services complained?

jakob1379:
I believe it was adminer

tiangolo:
Strange, as adminer doesn't take any secrets of its own, only the ones from Postgres :thinking: 

I would say, we could close this one and wait for someone else to bump into this (or you if it happens), so that I can replicate the issue and confirm this is needed.

PR #2026, opened 2025-12-04
Tab titles

whattheheckman:
Makes each page dynamically change it's name for good google analytics recognition


whattheheckman:
Accidentally added this to upstream instead of orgin sorry!

PR #2027, opened 2025-12-05
🐛 Ensure frontend waits for backend and redirect unauthenticated users

bfium:
- Add depends_on to frontend service to wait for backend healthcheck
- Add beforeLoad guard to dashboard route to redirect to login when not authenticated
- Prevents nginx 'host not found' errors and infinite loading spinner


tiangolo:
I'll pass on this for now. :coffee:

PR #2053, opened 2025-12-19
👷 Run Smokeshow always, even on test failures

YuriiMotov:
👷 Run Smokeshow always, even on test failures

Same as https://github.com/fastapi/fastapi/pull/14538


tiangolo (approved):
Thanks! 🙌

PR #2067, opened 2025-12-26
Fix MagicMock misuse in pre-start tests for Python 3.12+

sajalshh:
**Fix MagicMock misuse in pre-start tests**

**What changed:**
- Fixed incorrect use of `MagicMock(return_value=True)` in pre-start tests
- Replaced invalid `called_once_with` checks with `assert_called_once_with`
- Ensured tests actually validate behavior instead of being no-ops

**Why?**
- Python 3.12+ tightened MagicMock behavior
- Existing tests error out due to invalid MagicMock usage
- This makes the tests meaningful and future-proof


tiangolo:
It seems this was handled in another PR, so I'll close this one now. Thanks! :coffee:

PR #2182, opened 2026-02-12
🔧 Use uv default build system

patrick91:


patrick91:
Going to close this for now, as it is not really needed (yet)

PR #2184, opened 2026-02-13
♻️ Migrate to `DATABASE_URL` instead of separate variables

patrick91:


at146:
I don't know why you want to

atmosmaciel:
I understand that this is an important modification for abstracting (configuring) database connections, regardless of the engine chosen by the user.

patrick91:
@at146 sorry for the late reply, I was actually reconsidering this a bit

I usually migrate to a single env var for this, just for convenience (also most cloud's integrations expose that env variable, so it is easier to deploy)

> I understand that this is an important modification for abstracting (configuring) database connections, regardless of the engine chosen by the user.

you mean being able to do: `DATABASE_URL=postgres+engine://host/db`?

atmosmaciel:
> @at146 sorry for the late reply, I was actually reconsidering this a bit
> 
> I usually migrate to a single env var for this, just for convenience (also most cloud's integrations expose that env variable, so it is easier to deploy)
> 
> > I understand that this is an important modification for abstracting (configuring) database connections, regardless of the engine chosen by the user.
> 
> you mean being able to do: `DATABASE_URL=postgres+engine://host/db`?

Exactly. Using `DATABASE_URL` is more of a semantic and practical convenience than something necessarily more technical. It removes the need for a `SQLALCHEMY_DATABASE_URI` environment and multiple lines defining `POSTGRES_*`.

I don't know about all the impacts involved, but apparently it wouldn't be a big problem. It would be interesting if a maintainer could clarify some doubts about this.

@at146 any specific reason why this modification wouldn't be useful?

tiangolo (approved):
Nice, thank you! 🙌 

Sorry it took me so long to come back to this, even though I asked for it. 😅

Much simpler now. 🎉

PR #2236, opened 2026-03-17
fix: remove unused os import in alembic env.py

Sergy988:
Hi!
I noticed that the os module is imported at the top of backend/app/alembic/env.py but is never actually.

Removed the unused import to keep the file clean.


tiangolo:
Thanks! I prefer to keep this as is for now as this code is autogenerated by Alembic.

PR #2244, opened 2026-03-25
✨ Add clear error message for missing route tags in `app/main.py`

rxxbyy:
Added a guard statement with a clear error message when a route is missing tags. 

Currently, when you miss setting up a tag you get an error like:
```
return f"{route.tags[0]}-{route.name}"
          ~~~~~~~~~~^^^
IndexError: list index out of range
```

With the proposed changes, the developer sees:
```
ValueError: Route 'my_new_route' is missing a tag. Each route must have at least one tag for client generation.
```

I believe this improves developer experience and saves debugging time for anyone using the template.
Thank you!

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -8,6 +8,11 @@ from app.core.config import settings
 
 
 def custom_generate_unique_id(route: APIRoute) -> str:
+    if not route.tags:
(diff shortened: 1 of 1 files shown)


YuriiMotov (approved):
Makes sense!

PR #2253, opened 2026-04-05
♻️ Remove redundant item deletion logic in `backend/app/api/routes/users.py‎`

raphaeltannous:
Hey!

Removed redundant manual item deletion in the user deletion endpoint, since the `SQLModel` relationship already has `ON DELETE CASCADE` configured, which automatically deletes all related items when a user is deleted.

https://github.com/fastapi/full-stack-fastapi-template/blob/bba8d07c0cb4ac0e38a99d1de38090048fab8dee/backend/app/models.py#L93-L95


YuriiMotov (changes requested):
Makes sense!
But we should then add tests that check that items are also deleted.
Currently tests don't check that: https://github.com/fastapi/full-stack-fastapi-template/blob/master/backend/tests/api/routes/test_users.py

raphaeltannous:
> Makes sense! But we should then add tests that check that items are also deleted. Currently tests don't check that: https://github.com/fastapi/full-stack-fastapi-template/blob/master/backend/tests/api/routes/test_users.py

Hey! I've added 2 tests to check that the items are deleted.

backend/tests/api/routes/test_users.py
```diff
@@ -519,3 +520,64 @@ def test_delete_user_without_privileges(
...
+
+    r = client.delete(
+        f"{settings.API_V1_STR}/users/me",
+        headers=headers,
+    )
+    assert r.status_code == 200
+    deleted_user = r.json()
+    assert deleted_user["message"] == "User deleted successfully"
+    result = db.exec(select(User).where(User.id == user_id)).first()
+    assert result is None
```
YuriiMotov:
```suggestion
```

This is already tested above

backend/tests/api/routes/test_users.py
```diff
@@ -519,3 +520,64 @@ def test_delete_user_without_privileges(
...
+
+    r = client.delete(
+        f"{settings.API_V1_STR}/users/{user_id}",
+        headers=superuser_token_headers,
+    )
+    assert r.status_code == 200
+    deleted_user = r.json()
+    assert deleted_user["message"] == "User deleted successfully"
+    result = db.exec(select(User).where(User.id == user_id)).first()
+    assert result is None
```
YuriiMotov:
```suggestion
```

This is already tested above

YuriiMotov (dismissed):
LGTM!

I would simplify new added tests by excluding part that is already covered by previous tests

raphaeltannous:
> LGTM!
> 
> I would simplify new added tests by excluding part that is already covered by previous tests

Updated.

backend/tests/api/routes/test_users.py
```diff
@@ -519,3 +520,56 @@ def test_delete_user_without_privileges(
...
+
+def test_delete_user_me_deletes_items(client: TestClient, db: Session) -> None:
+    username = random_email()
+    password = random_lower_string()
+    user_in = UserCreate(email=username, password=password)
+    user = crud.create_user(session=db, user_create=user_in)
+    user_id = user.id
+
+    create_random_item(db)
+    create_random_item(db)
```
YuriiMotov: Looked one more time and noticed that this creates items not linked to the user created on the previous step.

We should modify `create_random_item` to accept optional `user_id` and use it if provided.
Also, let's add an assertion that user has items before deletion:

```
    items_query = select(Item).where(Item.owner_id == user_id)
    items_result = db.exec(items_query).all()
    assert len(items_result) > 0
```

YuriiMotov (changes requested):

raphaeltannous:
You are right! I did not notice that.

I created `create_random_item_for_user(db: Session, owner_id: uuid.UUID) -> Item`, and used it. I also added an assertion for the item creation.

backend/tests/utils/item.py
```diff
@@ -14,3 +16,11 @@ def create_random_item(db: Session) -> Item:
...
     item_in = ItemCreate(title=title, description=description)
     return crud.create_item(session=db, item_in=item_in, owner_id=owner_id)
+
+
+def create_random_item_for_user(db: Session, owner_id: uuid.UUID) -> Item:
+    assert owner_id is not None
+    title = random_lower_string()
+    description = random_lower_string()
+    item_in = ItemCreate(title=title, description=description)
+    return crud.create_item(session=db, item_in=item_in, owner_id=owner_id)
```
YuriiMotov: Why don't you just modify existing `create_random_item` like this?

```
def create_random_item(db: Session, owner_id: uuid.UUID | None = None) -> Item:
    if owner_id is None:
        user = create_random_user(db)
        owner_id = user.id
    assert owner_id is not None
    title = random_lower_string()
    description = random_lower_string()
    item_in = ItemCreate(title=title, description=description)
    return crud.create_item(session=db, item_in=item_in, owner_id=owner_id)
```

raphaeltannous: Thank you for pointing that out.

YuriiMotov (approved):
LGTM!

@raphaeltannous, thanks!
Passing this to Sebastian

PR #2260, opened 2026-04-16
🔒️ Add zizmor and fix audit findings

YuriiMotov:
Changes applied:

* Setup daily interval and 7 days cooldown period for Dependabot
* Added `pre-commit` package ecosystem to Dependabot config
* Ignored `dangerous-triggers` rule for `pull_request_target` and `workflow_run` (checked that they are used in a safe way)
* Specified minimal permissions on workflow level, moved permissions to the job level
* Ignored `secrets-outside-env` rule as using the environments would require approval for each run (and without required approvals it wouldn't make sense)
* **Specified `environment` for `deploy-staging.yml` and `deploy-production.yml` workflows to protect secrets, updated deployment docs**
* Added `persist-credentials: false` for `actions/checkout` when persisting is not needed by other steps
* Specified version of uv to install for `astral-sh/setup-uv` (Note that Dependabot will not upgrade it, but Renovate can do it)
* Specified run condition in latest-changes to make it clear that it only runs for merged PRs
* Replaced `uvx prek` command with `uv run prek` - `uvx` uses latest version (unpinned), it's better to use locked version
* Added zizmor pre-commit hook
* Added zizmor workflow to check on push to `master` (`pre-commit` hook will only work if workflow files updated by PR)
* Added timeouts for jobs (except `deploy-**` jobs as they are not being run or our repo and should probably be adjusted in real project repos) -  multiplied average time by 2-3, and set 5 minutes for short (less than 1 minute) tasks

---

Added smokeshow and zizmor dependencies to root level project as I think they should be there.
I think we should also move `prek` there and maybe other dev dependencies


.github/dependabot.yml
```diff
@@ -5,22 +5,28 @@ updates:
...
+      default-days: 7
     commit-message:
       prefix: ⬆
     labels: [dependencies, internal]
   # Python uv
   - package-ecosystem: uv
     directory: /
     schedule:
-      interval: weekly
+      interval: daily
```
YuriiMotov:
```suggestion
      interval: weekly
```

.github/dependabot.yml
```diff
@@ -5,22 +5,28 @@ updates:
...
+      default-days: 7
     commit-message:
       prefix: ⬆
     labels: [dependencies, internal]
   # bun
   - package-ecosystem: bun
     directory: /
     schedule:
-      interval: weekly
+      interval: daily
```
YuriiMotov:
```suggestion
      interval: weekly
```

.github/dependabot.yml
```diff
@@ -32,15 +38,28 @@ updates:
       - /backend
       - /frontend
     schedule:
-      interval: weekly
+      interval: daily
```
YuriiMotov:
```suggestion
      interval: weekly
```

.github/dependabot.yml
```diff
@@ -32,15 +38,28 @@ updates:
...
+      default-days: 7
     commit-message:
       prefix: ⬆
     labels: [dependencies, internal]
   # Docker Compose
   - package-ecosystem: docker-compose
     directory: /
     schedule:
-      interval: weekly
+      interval: daily
```
YuriiMotov:
```suggestion
      interval: weekly
```

tiangolo (approved):
Excellent, thank you! 🚀

PR #2285, opened 2026-05-15
🐛 Fix mobile sidebar logout leaving login page unclickable

Haohua-Sun:
Fixes a mobile sidebar interaction issue where logging out from the user menu can leave the login page visible but not clickable.

The logout navigation could run before the nested dropdown menu and mobile sidebar finished closing, leaving dialog/dropdown interaction state active after navigation. The fix closes the dropdown first, then waits for the mobile sidebar close animation before logging out.

Tested by reproducing the issue before the fix and verifying it after the fix on Chrome and Edge on Windows, Android emulator on Windows, and Safari on macOS.


Haohua-Sun:
The issue is no longer reproducible with the latest dependencies. The underlying Radix cleanup bug has been fixed upstream, so I’m closing this PR as no longer needed.

PR #2310, opened 2026-05-26
📝 Simplify error message for non-existent user in password recovery

Musbi8788:
# This PR improve the error message on the  `login.py` for better UX

## Updated the response message from:

- "The user with this username does not exist in the system."
to:
- "User does not exist"

This  makes the reponse shoter and remove unnessary message like ".... username don't exit in the system" for better security with telling fake users that the name don't exist in the system instead we can just tell them the user does not exist.

## Checklist

- [ ] This PR is an obvious typo fix, or it links to a GitHub Discussion for the proposed code change.
- [ ] Coverage stays at 100%.

### DIFF
--- a/backend/app/api/routes/login.py
+++ b/backend/app/api/routes/login.py
@@ -111,7 +111,7 @@ def recover_password_html_content(email: str, session: SessionDep) -> Any:
     if not user:
         raise HTTPException(
             status_code=404,
-            detail="The user with this username does not exist in the system.",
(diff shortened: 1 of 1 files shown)


Musbi8788:
> This was marked as potentially AI generated and will be closed now. If this is an error, please provide additional details, make sure to read the docs about contributing and AI.

Thank you for this update am updating the PR

PR #2318, opened 2026-05-29
📝 Simplify error message for non-existent user in password recovery

Musbi8788:
## Pull Request
Simplify error message 

## Description
This makes the reponse shoter and remove unnessary message like ".... username don't exit in the system" for better security with telling fake users that the name don't exist in the system instead we can just tell them the user does not exist.

Updated the response message from:
"The user with this username does not exist in the system."
to:
"User does not exist"

## AI Disclaimer

No AI used

## Checklist

- [x] This PR is an obvious typo fix, or it links to a GitHub Discussion for the proposed code change.
- [ ] Coverage stays at 100%.

### DIFF
--- a/backend/app/api/routes/login.py
+++ b/backend/app/api/routes/login.py
@@ -111,7 +111,7 @@ def recover_password_html_content(email: str, session: SessionDep) -> Any:
     if not user:
         raise HTTPException(
             status_code=404,
-            detail="The user with this username does not exist in the system.",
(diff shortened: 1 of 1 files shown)


alejsdev:
Hi @Musbi8788, we don't think this change is necessary, so we'll be closing it again. Thank you for your understanding.

Musbi8788:
Thank you

PR #2370, opened 2026-07-03
Keep password recovery generic when email delivery is unavailable

DivyamTalwar:
## Summary

This keeps `/password-recovery/{email}` on the same generic HTTP response when reset email delivery is unavailable.

The route now:

- returns the generic password-recovery response without looking up a user when email delivery is not configured
- only generates and sends a reset email when delivery is configured and the user exists
- returns the same generic response if reset-token generation, email-template generation, or `send_email()` raises

## Why

The endpoint already returns a generic response for unknown emails. Before this change, an existing user could still hit an internal error when SMTP settings were disabled because `send_email()` requires email delivery settings to be configured.

This avoids status/body-based differences for disabled-email and reset-email failure paths. It does not try to equalize timing between existing and non-existing users, and it does not guarantee that an email was delivered.

## Tests

- `cd backend && uv run --no-sync ruff check app/api/routes/login.py tests/api/routes/test_login.py` — passed
- `cd backend && uv run --no-sync ruff format --check app/api/routes/login.py tests/api/routes/test_login.py` — passed
- `cd backend && uv run --no-sync pytest tests/api/routes/test_login.py --collect-only -q` — passed, collected 13 tests
- `cd backend && uv run --no-sync bash scripts/prestart.sh && uv run --no-sync pytest tests/api/routes/test_login.py -q` — passed, 13 tests passed with a temporary local `postgres:18-alpine` service

I did not complete the full GitHub Actions-equivalent stack locally; CI should cover the Compose/mailcatcher/coverage/Playwright surfaces.

## Notes

No existing issue. I checked the current open upstream issues and PRs for password recovery / SMTP / disabled email / `send_email` duplicates and did not find a matching open item.

### DIFF
--- a/backend/app/api/routes/login.py
+++ b/backend/app/api/routes/login.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 from typing import Annotated, Any
 
@@ -18,6 +19,7 @@ from app.utils import (
 )
 
 router = APIRouter(tags=["login"])
+logger = logging.getLogger(__name__)
 
 
 @router.post("/login/access-token")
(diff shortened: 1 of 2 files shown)


DivyamTalwar:
Thanks! The code/test checks are green; the remaining failed check is the repository label gate requiring one of the release labels. This looks like a bug-fix PR, but I do not have permission to apply labels to the upstream PR. A maintainer can unblock that check by adding the `bug` label if that classification looks right.

PR #2383, opened 2026-07-16
feat: add pagination limits to user and item list endpoints

HassanRady:
## Pull Request

Discussion: 

## Description

Currently, pagination routes accept an arbitrary `limit` parameter without validation (e.g., `limit: int = 100`). At scale, a client or malicious user could supply an extremely large limit (e.g., `?limit=1000000`). 
FastAPI and SQLModel will read all matching records from the database and instantiate them as Python objects in the server's heap memory. This results in:
1. **Out-of-Memory (OOM) Crashes**: The container's memory balloons, triggering the kernel OOM-killer and terminating the backend worker.
2. **CPU Starvation**: Heavy garbage collection cycles block other concurrent requests on the event loop.

### The Remediation Code
Use FastAPI's `Query` parameter validator to enforce a hard maximum limit (`le=100`). This rejects invalid requests with a `422 Unprocessable Entity` validation error before querying the database.

## Checklist

- [ ] This PR links to a GitHub Discussion for the proposed code change.
- [x ] The new or updated tests fail on the main branch and pass on this PR.
- [ ] Coverage stays at 100%.

### DIFF
--- a/backend/app/api/routes/items.py
+++ b/backend/app/api/routes/items.py
@@ -1,7 +1,7 @@
 import uuid
 from typing import Any
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, HTTPException, Query
 from sqlmodel import col, func, select
 
 from app.api.deps import CurrentUser, SessionDep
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for the interest. For now, I'll leave this as an implementation detail to be decided by each app. :coffee:

PR #2384, opened 2026-07-18
🐛 Add background to the sticky layout header

istoutjesdijk:
The header is sticky (sticky top-0) but does not set a background, so page content is visible through it when you scroll.


alejsdev:
Good catch, thank you @istoutjesdijk! 🎉

PR #2447, opened 2026-08-21
fix(backend): mount frontend route only when FRONTEND_DIR exists

aoright:
### Summary
- Check if `FRONTEND_DIR` exists before calling `app.frontend("/", directory=FRONTEND_DIR)` in `backend/app/main.py`.
- Prevents `RuntimeError` during local backend development or test collection when the frontend build directory is not yet generated or mounted.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -33,4 +33,5 @@ app.add_middleware(
 )
 
 app.include_router(api_router, prefix=settings.API_V1_STR)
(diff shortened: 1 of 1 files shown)


YuriiMotov:
You can avoid having `RuntimeError` by using `fatsapi dev` command.

You can also specify `FASTAPI_ENV: "development"` env variable and use `fastapi run` (see [FastAPI docs](https://fastapi.tiangolo.com/tutorial/frontend/#check-directory))

PR #2449, opened 2026-08-21
Add rbac

dehbok:
Summary

Replaces the boolean is_superuser flag with role-based access control (RBAC): three roles (admin, manager, member), five permission codes, and a single enforcement path on both backend and frontend, so future privileged features are one dependency call instead of another superuser check.

What changed

Data model & migration
- New role, permission, role_permission tables; user.role_id replaces user.is_superuser
- Migration 2d1ae9889b29 seeds the three roles and five permission codes and backfills existing users (is_superuser=true → admin, else → member)
- backend/app/core/rbac.py is the single source of truth for permission codes and role→permission grants (ROLE_PERMISSIONS)

Backend enforcement
- New require_permission(code) dependency (app/api/deps.py), applied per-route — no middleware, no decorators, no inline role == checks
- Permission remap: users:list, users:create, users:manage, metrics:view, and a system:admin catch-all (item cross-owner access, transactional-email endpoints, self-delete guard)
- Item-ownership fallback stays inline since it's data-scoped, not role-scoped
- New stub GET /api/v1/metrics/ endpoint gated on metrics:view
- init_db upserts seed data on every startup so DB state always matches rbac.py

Frontend
- Regenerated OpenAPI client (metrics route, role/permissions replace is_superuser)
- New hasPermission(user, code) helper mirroring the backend's five codes — no component branches on user.role
- New shared AccessDenied page at /forbidden; every guard redirects there instead of bouncing to /
- Sidebar gating on permission codes; new "Metrics" nav item and page
- Add/Edit User forms: two-way admin toggle → three-way role select
- Settings danger-zone gated on system:admin

Docs — new docs/AUTHORIZATION.md (permission matrix, mechanics, setup/seed/test instructions), one-line pointer from README.md

Tests — pytest fixtures updated + new test_rbac.py; Playwright specs updated + new Access Denied assertion

Housekeeping — .gitignore additions, a real mypy fix in crud.py, formatting cleanup, removed two dead doc links

Breaking: is_superuser is gone from user payloads/schemas — consumers must switch to role and the new permissions array on /users/me.


alejsdev:
Hello @dehbok 👋🏻 

Thanks for your interest in the project. For this kind of new feature, we normally prefer to start with a discussion before jumping into the implementation. We already have RBAC support for the template on our roadmap, so there's some overlap with work we're planning on our side. For next time, before spending too much time implementing a new feature, please check with someone on the team first to make sure we're aligned. Given this, I'll close this PR but thank you for the effort.

PR #2455, opened 2026-08-22
Fix API bugs surfaced by live tracing (422/400 instead of 500)

Vinv-AI:
## Pull Request

Discussion: https://github.com/fastapi/full-stack-fastapi-template/discussions/2454

## Description

Several endpoints returned 500 (or crashed) on inputs that should be rejected
with a 4xx. This PR fixes them and adds tests for each case.

- **Pagination (items, users):** `skip`/`limit` were unbounded, so negative
  values reached Postgres `OFFSET`/`LIMIT` and 500'd. Added shared `SkipQuery`
  (`ge=0`) and `LimitQuery` (`ge=1, le=100`) in `app/api/params.py`; out-of-range
  values now return 422.
- **Private user create (dev-only `/private` router):** `email` was a bare `str`,
  so an invalid address was committed and every later read 500'd on `UserPublic`
  (`EmailStr`). Now uses `EmailStr` (`max_length=255`) and password bounds
  (`8–128`) → 422 before insert; drops the accepted-but-ignored `is_verified`
  (no such column).
- **Private user create:** a duplicate email raised an unhandled `UniqueViolation`
  → 500. Now looks up first and returns 400; creates via `crud.create_user`.
- **Password recovery HTML endpoint:** the response header name was `"subject:"`,
  an invalid HTTP header name that raised `RuntimeError`. Now `"subject"`.
- **Password recovery:** the email is sent via `BackgroundTasks` so the HTTP
  response no longer waits on SMTP (the enumeration-safe uniform response is
  unchanged).

No public/documented API changes. B1–B3 are on the `private` router, which is
only mounted when `FASTAPI_ENV == "development"` (E2E/local).

## AI Disclaimer

Models: Composer 2.5 (via Cursor + VinvAI)

Prompt (paraphrased): apply the live-trace fix set (unbounded pagination → 422,
private-route email/password validation and duplicate handling, the invalid
`subject:` recovery header, and background-task SMTP) — verifying each bug
against the current code first — and add a test for each case.

<details>
<summary>AI transcript</summary>

- Verified each bug against the current code (`private.py` bare `email: str` +
  no dup check; `users.py`/`items.py` unbounded `skip`/`limit`; `login.py`
  `"subject:"` header and synchronous `send_email`; confirmed `is_verified` has
  no model column and the `private` router is dev-only).
- Implemented: `app/api/params.py` (`SkipQuery`/`LimitQuery`); `users.py` /
  `items.py` use them; `private.py` (`EmailStr` + bounds, duplicate → 400,
  `crud.create_user`); `login.py` (`subject` header, `BackgroundTasks`).
- Added tests: invalid email, short password, duplicate email (private);
  negative skip (users); negative limit (items); valid recovery header,
  background-vs-no-op email (login).
- Dropped a proposed change to the superuser-only recovery-HTML 404 (no real
  enumeration benefit; returning a preview for a non-existent user is wrong).
- Verified locally: all changed files compile; `ruff check` and `ruff format`
  are clean. The pytest suite was NOT run locally (no Postgres available in this
  environment) — see checklist note.

</details>

## Checklist

- [x] This PR links to a GitHub Discussion for the proposed code change.
- [x] I added tests for the change.
- [ ] Coverage stays at 100%. 

### DIFF
--- a/backend/app/api/params.py
+++ b/backend/app/api/params.py
@@ -1,0 +1,9 @@
+from typing import Annotated
+
+from fastapi import Query
+
+# Shared pagination query parameters. Bounds are enforced at the API layer so
+# invalid values return 422 instead of reaching Postgres OFFSET/LIMIT (which
+# rejects negative values and would otherwise surface as a 500).
+SkipQuery = Annotated[int, Query(ge=0)]
+LimitQuery = Annotated[int, Query(ge=1, le=100)]
--- a/backend/app/api/routes/items.py
+++ b/backend/app/api/routes/items.py
@@ -5,6 +5,7 @@ from fastapi import APIRouter, HTTPException
 from sqlmodel import col, func, select
 
 from app.api.deps import CurrentUser, SessionDep
+from app.api.params import LimitQuery, SkipQuery
 from app.models import Item, ItemCreate, ItemPublic, ItemsPublic, ItemUpdate, Message
 
 router = APIRouter(prefix="/items", tags=["items"])
@@ -12,7 +13,10 @@ router = APIRouter(prefix="/items", tags=["items"])
 
 @router.get("/", response_model=ItemsPublic)
 def read_items(
-    session: SessionDep, current_user: CurrentUser, skip: int = 0, limit: int = 100
+    session: SessionDep,
+    current_user: CurrentUser,
+    skip: SkipQuery = 0,
+    limit: LimitQuery = 100,
 ) -> Any:
     """
     Retrieve items.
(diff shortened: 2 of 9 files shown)


YuriiMotov:
Closing this for the following reasons:

1. We ask to only open PRs when it's explicitly requested by maintainers. See https://tiangolo.com/open-source/contributing/ and [PR description template](https://github.com/fastapi/.github/blob/main/.github/pull_request_template.md?plain=1)

2. This looks like an AI-generated PR without proper human review ([see](https://tiangolo.com/open-source/contributing/#closing-automated-and-ai-prs))

3. PRs should be focused. This one solves several unrelated issues, and it makes it harder to review

PR #2457, opened 2026-08-24
Updated data initialization to include try catch statements

Alien-Axolotl:
## Summary

* Added a `try...except` block to `initial_data.py` to catch and log any errors that occur during initial data creation.
* Re-raises the exception so the script properly exits with a non-zero status code.


YuriiMotov:
Thanks, but I think we don't need this. The Script already fails loudly in case of error

PR #2474, opened 2026-09-02
🐛 Fix return 401 instead of 404 when a token's user doesn't exist

ErikH-move:
get_current_user returned 404 when the user id it encoded no longer existed. 404 is the wrong code since nothing about the requested resource is missing, the presented credential is just invalid. An issue with the 404 code was that the frontend's global error handler only clears and redirects to login on 401/403.


YuriiMotov:
I agree 401 is correct status code here.

@ErikH-move, could you please add a test?

YuriiMotov (approved):
LGTM!

PR #2484, opened 2026-09-19
fix: parenthesize multiple exceptions in deps.py to fix SyntaxError on Python <= 3.13

H-XX-D:
### Problem
In `backend/app/api/deps.py`, line 36 catches multiple exceptions without enclosing parentheses:
```python
except InvalidTokenError, ValidationError:
```
On Python 3 (including Python 3.10, 3.11, 3.12, and 3.13), running `python3 -m py_compile backend/app/api/deps.py` fails at parse time:
```text
  File "backend/app/api/deps.py", line 36
    except InvalidTokenError, ValidationError:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized
```

### Solution
Parenthesize the exception types to `except (InvalidTokenError, ValidationError):`.

### Verification
Verified with `python3 -m py_compile backend/app/api/deps.py` (exits 0).


H-XX-D:
Closing after checking the repository's supported runtime. [backend/pyproject.toml](https://github.com/fastapi/full-stack-fastapi-template/blob/cb740b656d7a0a6c5e12c7bf8e50343ec94ee9c7/backend/pyproject.toml) already requires `>=3.14,<4.0` and sets Ruff's target to `py314`.

The original `except InvalidTokenError, ValidationError:` syntax is valid in Python 3.14 under [PEP 758](https://peps.python.org/pep-0758/), and I confirmed that the unchanged file compiles with CPython 3.14.6. The reported failure occurs on Python 3.13, outside the supported range.

The formatting commit restored the original syntax, so this PR also has no remaining diff. This was a false positive for the project's supported runtime. Sorry for the noise.

